PR #2, opened 2024-02-28
CDP-22: Add the organisation-app to docker compose services

jakzal:
The Dockerfile is constructed from multiple stages:

* base - installs all system dependencies
  * dependencies - installs application dependencies
  * source - bundles application source code
    * build - packages the application as static files
    * dev - runs the application in development mode
  * serve - serves the application packages as static files
 
Docker compose mounts the `Frontend/organisation-app` as a volume, so it's possible to use this setup for development.

## Developer notes

If you have the `compose.override.yml` file locally, you'll need to either remove it so it's re-generated, or add the following section under `services`:

```
  organisation-app:
    ports:
      - "${CDP_ORGANISATION_APP_PORT:-3000}:3000"
```

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -15,8 +15,8 @@
 **/bin
 **/charts
 **/docker-compose*
+**/compose*
 **/Dockerfile*
-**/node_modules
 **/npm-debug.log
 **/obj
 **/secrets.dev.yaml
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -41,6 +41,7 @@ jobs:
     runs-on: ubuntu-latest
     name: Build containers
     env:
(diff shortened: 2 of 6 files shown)


.dockerignore
```diff
@@ -15,8 +15,8 @@
 **/bin
 **/charts
 **/docker-compose*
+**/compose*
 **/Dockerfile*
-**/node_modules
```
jakzal, line 19: This could make docker context quite bigger, but ignoring node_modules will trigger installation every time we start the container.

glgoaco (approved):

.dockerignore
jakzal: Fixed in #6

PR #12, opened 2024-03-11
CDP-10 Prototype for OneLogin Client

glgoaco:


.gitignore
```diff
@@ -45,3 +45,6 @@ msbuild.log
 msbuild.err
 msbuild.wrn
 OpenAPI
+
+appsettings.json
+appsettings.*.json
```
jakzal, line 50: This should probably go to the project's gitignore instead of the global one. Otherwise it affects files that are already committed to the repo.

glgoaco: Done ✅

PR #18, opened 2024-03-15
Introduced session in MVC App

dharmverma:
Introduced session in MVC App

### DIFF
--- a/Frontend/CO.CDP.OrganisationApp/Session.cs
+++ b/Frontend/CO.CDP.OrganisationApp/Session.cs
@@ -6,7 +6,7 @@ namespace CO.CDP.OrganisationApp
     {
         public T? Get<T>(string key)
         {
-            CheckSessionIsNull(httpContextAccessor.HttpContext);
+            CheckSessionIsNull();
 
             var value = httpContextAccessor.HttpContext!.Session.GetString(key);
             return value == null ? default : JsonSerializer.Deserialize<T>(value);
@@ -14,7 +14,7 @@ namespace CO.CDP.OrganisationApp
 
         public void Set<T>(string key, T value)
         {
-            CheckSessionIsNull(httpContextAccessor.HttpContext);
+            CheckSessionIsNull();
 
             httpContextAccessor.HttpContext!.Session
                 .SetString(key, JsonSerializer.Serialize(value));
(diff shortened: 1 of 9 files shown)


Frontend/CO.CDP.OrganisationApp/Session.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System.Text.Json;
+
+namespace CO.CDP.OrganisationApp
+{
+    public class Session(IHttpContextAccessor httpContextAccessor) : ISession
```
glgoaco, line 5: All methods require the httpContextAccessor, it makes sense to set it in the constructor.

` private readonly IHttpContextAccessor _httpContextAccessor;

        public Session(IHttpContextAccessor httpContextAccessor)
        {
            _httpContextAccessor = httpContextAccessor;
        }`

Frontend/CO.CDP.OrganisationApp/Session.cs
```diff
@@ -0,0 +1,38 @@
...
+        }
+
+        public void Remove(string key)
+        {
+            CheckSessionIsNull(httpContextAccessor.HttpContext);
+
+            httpContextAccessor.HttpContext!.Session.Remove(key);
+        }
+
+        private static void CheckSessionIsNull(HttpContext? context)
```
glgoaco, line 30: This would simplify the code as we do not need to pass the context each time.

`private void CheckSessionIsNull()
        {
            if (_httpContextAccessor.HttpContext?.Session == null)
            {
                throw new Exception("Session is not available");
            }
        }`

glgoaco (approved):
LGTM

Frontend/CO.CDP.OrganisationApp/Session.cs
dharmverma: It is set in the constructor @glgoaco  
C# 12 introduces [primary constructors](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/instance-constructors#primary-constructors), a concise syntax to declare constructors whose parameters are available anywhere in the body of the type.

Frontend/CO.CDP.OrganisationApp/Session.cs
dharmverma: Done 👍

JBaigGoaco (approved):

PR #19, opened 2024-03-18
Refactor session tests to avoid mocking a 3rd party type

jakzal:
Here's how the Session can be tested without mocking the 3rd party type, for reasons explained in https://github.com/cabinetoffice/GCGS-Central-Digital-Platform/pull/18#issuecomment-1999813018.

While refactoring I found an additional case where the HTTP context might be null, which beautifully proves the point.

### DIFF
--- a/Frontend/CO.CDP.OrganisationApp.Tests/SessionTest.cs
+++ b/Frontend/CO.CDP.OrganisationApp.Tests/SessionTest.cs
@@ -1,27 +1,33 @@
-using System.Text;
-using System.Text.Json;
 using FluentAssertions;
 using Microsoft.AspNetCore.Http;
-using Moq;
-using HttpSession = Microsoft.AspNetCore.Http.ISession;
+using Microsoft.AspNetCore.Session;
+using Microsoft.Extensions.Caching.Distributed;
+using Microsoft.Extensions.Caching.Memory;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
 
 namespace CO.CDP.OrganisationApp.Tests
 {
     public class SessionTest
     {
-        const string dummySessionKey = "session_key";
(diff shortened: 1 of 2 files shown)


Frontend/CO.CDP.OrganisationApp/Session.cs
```diff
@@ -6,32 +6,33 @@ public class Session(IHttpContextAccessor httpContextAccessor) : ISession
     {
         public T? Get<T>(string key)
         {
-            CheckSessionIsNull();
-
-            var value = httpContextAccessor.HttpContext!.Session.GetString(key);
+            var value = GetSession().GetString(key);
```
jakzal, line 9: This is a bonus change. 

Asking for a validated value is less error-prone and should be easier to read than a validation method followed by the access to validated value.

Previously, `CheckSessionIsNull` validated if `httpContextAccessor.HttpContext!.Session` can be accessed, and then the `Get` method accessed the value. These two lines of code are of different levels of abstraction.

Frontend/CO.CDP.OrganisationApp.Tests/SessionTest.cs
```diff
@@ -1,63 +1,70 @@
...
+using Microsoft.Extensions.Caching.Distributed;
+using Microsoft.Extensions.Caching.Memory;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
 
 namespace CO.CDP.OrganisationApp.Tests
 {
     public class SessionTest
     {
-        const string dummySessionKey = "session_key";
```
jakzal, line 12: I inlined the key and value, as it's relevant in the test cases.

glgoaco (approved):

dharmverma (approved):

JBaigGoaco (approved):

PR #25, opened 2024-03-19
DP-6 Supplier registers as a Tenant - backend

jakzal:
DP-6 tasks covered by this PR:
* DP-16: Create the TenantApi Endpoint to handle POST /tenants
* DP-17: Support GET /tenants/{tenantId} in TenantApi
* DP-19: Add a docker container with postgresql database to the developer setup

## Summary of changes

* `POST /tenants` and `GET /tenants/{tenantId}` endpoints are implemented.
* `db` service (PostgreSQL) is now included in Docker compose configuration.
*  FluentAssertions are used instead of regular xunit assertions.
* `DatabaseTenantRepository` throws custom exceptions instead of the low-level ones in case of validation errors.
* Use Case layer is placed between the Web Api and Persistence to handle application logic.
* The Web Api layer is tested with HTTP client, but use cases that it calls are mocked.
* AutoMapper is used to map between the Web Api model and the Persistence model.
* AutoMapper mappings are tested as part of use case tests to avoid adding another layer of indirection.
* Web Api model classes are move to its own `Model` namespace (for now only those that were needed for the current feature). Types that are meant to modify the state were made Commands, while the type names exposed via the OpenAPI contract were kept as before. 
* Diagrams were updated, but they're still work in progress.

Component diagram for the Tenant API

## What's still missing

* Error handling

## db service

Once merged, the following entry needs to be added to `compose.override.yml` to map the port for the database:

```yaml
services:
  db:
    ports:
      - "${CDP_DB_PORT:-5432}:5432"
    deploy:
      replicas: 1
```


Services/CO.CDP.Tenant.WebApi/appsettings.Development.json
```diff
@@ -6,6 +6,6 @@
     }
   },
   "ConnectionStrings": {
-    "TenantDatabase": "Server=localhost;Database=cdp;Username=cdp;Password=cdp;"
+    "TenantDatabase": "Server=db;Database=cdp;Username=cdp_user;Password=cdp123;"
```
glgoaco, line 9: Should we block all app setting files with the exception of appsettings-template.json ?

glgoaco (approved):

Services/CO.CDP.Tenant.WebApi/appsettings.Development.json
jakzal: I'd prefer to have it working without any manual steps required (like having to copy the template). Settings like database connection will be overridden with environment variables when deployed to AWS.

dharmverma (approved):

Services/CO.CDP.Tenant.WebApi/appsettings.Development.json
jakzal: @glgoaco for example, to override the connection string, the following environment variable needs to be defined:

```
ConnectionStrings__TenantDatabase="Server=localhost;Database=MyDb;Username=MYUSERPassword=MYPASS;"
```

I actually added it to the HTTP launch profile now, so that if we run the tenant service from the IDE it connects to the DB that's run on localhost (docker).

https://github.com/cabinetoffice/GCGS-Central-Digital-Platform/blob/8d0e255321b88c2694d77ecfa8083b759d25b4e1/Services/CO.CDP.Tenant.WebApi/Properties/launchSettings.json#L20

PR #32, opened 2024-03-22
Draft the ITenantClient interface

jakzal:

PR #35, opened 2024-03-25
Decide how to create webapi clients

jakzal:
I considered three options and put forward my recomendation. Please give feedback and let me know if there's anything I haven't considered.

One option for client generation I didn't explore is https://github.com/OpenAPITools/openapi-generator.

For better reviewing experience, GitHub can render the document on the PR when you "view the file": https://github.com/cabinetoffice/GCGS-Central-Digital-Platform/blob/6ffc419b3276dc66fe46d9705fe1254d631d3b99/docs/decisions/0002-create-webapi-clients.adoc

### DIFF
--- a/docs/decisions/0002-create-webapi-clients.adoc
+++ b/docs/decisions/0002-create-webapi-clients.adoc
@@ -1,0 +1,161 @@
+= 2. Create web api clients
+
+Date: 2023-03-25
+
+== Status
+
+Draft
+
+== Context
+
+We need to create clients for each of our web api services to be used internally (at least initially).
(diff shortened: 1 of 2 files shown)


dharmverma (approved):

glgoaco:
There is a risk of Microsoft deprecating NSwag in the future, but I feel it is the right tool for now as it is mature, ticks the box in terms of features and will save us time.

jakzal:
> There is a risk of Microsoft deprecating NSwag in the future, but I feel it is the right tool for now as it is mature, ticks the box in terms of features and will save us time.

Indeed. Furthermore, the interface that NSwag generates is quite close to what we would've defined anyway (#32). This decision can be rather easily reversed.

PR #52, opened 2024-03-28
Choose the database migration strategy

jakzal:
Best viewed when rendered by GitHub: https://github.com/cabinetoffice/GCGS-Central-Digital-Platform/blob/bc80a0b5e9a140e366d284b013e9b9c0faa06f23/docs/decisions/0003-applying-database-migrations.adoc


dpatel017 (approved):

dharmverma (approved):
I like the idea of option 5 - self containerised migration bundle, although never used before, but happy to learn and explore.

JBaigGoaco (approved):

PR #56, opened 2024-04-09
Supplier login tenant lookup before registration

dharmverma:


dpatel017 (approved):

JBaigGoaco (approved):

Frontend/CO.CDP.OrganisationApp/Pages/Registration/OneLoginCallback.cshtml.cs
```diff
@@ -16,28 +16,34 @@ public async Task<IActionResult> OnGet()
...
+            tenant = await tenantClient.LookupTenantAsync(userInfo.UserId);
         }
-        catch (ApiException)
+        catch (ApiException ex) when (ex.StatusCode == 404)
         {
-            // show error page?
-            throw new Exception("Unable to create tenant");
+            try
+            {
+                tenant = await tenantClient.CreateTenantAsync(
```
jakzal, line 29: I think we should first query the tenant before we attempt to create it. Querying for a tenant is not an exceptional path. Furthermore, people will log in far more often than they register. Trying to register them every time triggers unnecessary calls.

jakzal (changes requested):

Frontend/CO.CDP.OrganisationApp/Pages/Registration/OneLoginCallback.cshtml.cs
dharmverma: We are doing tenant lookup already here @jakzal, are you suggesting doing a get also?

jakzal: Ignore me. Looking at the diff it looked like the order was reversed (create, then lookup). Now that I just opened that file it's ok!

jakzal (approved):

PR #60, opened 2024-04-12
Choose error handing & data validation strategy

glgoaco:
Best viewed when rendered by GitHub: https://github.com/cabinetoffice/GCGS-Central-Digital-Platform/blob/error-handing-and-validation-decisions/docs/decisions/0004-error-handing-and-validation.adoc


Services/CO.CDP.Person.WebApi/Api/ValidationOption1.cs
```diff
@@ -0,0 +1,78 @@
...
+                Id = index.ToString(),
+                Name = $"Sussan Tables {index}",
+                Age = 40 + index,
+                Email = "sussan@example.com"
+            });
+
+        [HttpPost]
+        public IActionResult CreatePerson([FromBody] NewPerson newPerson)
+        {
+            if (!ModelState.IsValid)
```
jakzal, line 51: I'm aware it's only an example, but I wanted to clarify how we'd do it in practice.

I think the controller's responsibility is to:

* parse the request to construct the command
* pass the command to the service or model
* interpret the result
* construct the response

Any validation done in the web layer should be related to the HTTP message (most often its body payload) to confirm we can accept the request.

Model validation should be done in layers called by the web layer. The result should then be returned back from the service or model layer in a form of a return value or an exception.

Services/CO.CDP.Person.WebApi/Api/ValidationOption2.cs
```diff
@@ -0,0 +1,165 @@
...
+                {
+                    operation.OperationId = "ListPersons";
+                    operation.Description = "A list of persons.";
+                    operation.Summary = "A list of persons.";
+                    operation.Responses["200"].Description = "A list of persons.";
+                    return operation;
+                });
+            app.MapPost("/persons", async (HttpContext http, IValidator<NewPerson> validator, [FromBody] NewPerson newPerson) =>
+                {
+                    FluentValidation.Results.ValidationResult results = await validator.ValidateAsync(newPerson);
```
jakzal, line 57: What's good about this is we can have multiple errors returned in a single validation pass, while throwing an exception would only allow for one error.

dharmverma (approved):

dpatel017 (approved):

PR #77, opened 2024-05-02
Bugs/dp 121 refactor organisation model new

JBaigGoaco:
Bugs/dp 121 refactor organisation model new

### DIFF
--- a/Services/CO.CDP.Organisation.WebApi/Model/Command.cs
+++ b/Services/CO.CDP.Organisation.WebApi/Model/Command.cs
@@ -2,43 +2,37 @@ using System.ComponentModel.DataAnnotations;
 
 namespace CO.CDP.Organisation.WebApi.Model;
 
-public record RegisterOrganisation
+public record NewOrganisation
 {
     [Required(AllowEmptyStrings = true)] public required string Name { get; init; }
 
-    [Required]
     public required OrganisationIdentifier Identifier { get; init; }
 
-    [Required]
-    public required List<OrganisationIdentifier> AdditionalIdentifiers { get; init; }
+    public List<OrganisationIdentifier>? AdditionalIdentifiers { get; init; }
 
     [Required]
     public required OrganisationAddress Address { get; init; }
 
-    [Required]
-    public required OrganisationContactPoint ContactPoint { get; init; }
+    public required OrganisationContactPoint? ContactPoint { get; init; }
 
-    [Required]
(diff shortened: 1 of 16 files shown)


dharmverma (approved):

dpatel017 (approved):

Services/CO.CDP.Organisation.WebApi/Model/Command.cs
```diff
@@ -2,43 +2,37 @@
 
 namespace CO.CDP.Organisation.WebApi.Model;
 
-public record RegisterOrganisation
+public record NewOrganisation
```
jakzal, line 5: This was called "RegisterOrganisation" as it was treated as a command. It's still in the Command namespace, so has to either follow command naming conventions (an intended action to invoke), or the namespace should change.

Services/CO.CDP.Organisation.WebApi/Model/Command.cs
```diff
@@ -2,43 +2,37 @@
...
+    [Required]
+    public required string Name { get; init; }
 
-    [Required(AllowEmptyStrings = true)]
+    [Required]
     public required OrganisationIdentifier Identifier { get; init; }
 
-    [Required]
-    public required List<OrganisationIdentifier> AdditionalIdentifiers { get; init; }
+    public List<OrganisationIdentifier>? AdditionalIdentifiers { get; init; }
```
jakzal, line 30: It's a list and it can be empty. There shouldn't be a need to make it nullable.

Services/CO.CDP.Organisation.WebApi/Model/Command.cs
```diff
@@ -2,43 +2,37 @@
...
     [Required]
     public required OrganisationAddress Address { get; init; }
 
-    [Required]
-    public required OrganisationContactPoint ContactPoint { get; init; }
+    public OrganisationContactPoint? ContactPoint { get; init; }
 
-    [Required]
-    public required List<int> Roles { get; init; }
+    public List<int>? Types { get; init; }
```
jakzal, line 37: Again, a list can be empty. No need to make it nullable.

Services/CO.CDP.Organisation.WebApi/Model/Command.cs
jakzal: If you want to move away from the command pattern, that's fine, but that requires namespace change.

PR #80, opened 2024-05-03
DP-65 Person is associated with the organisation in the context of the tenant (Web Api)

glgoaco:


Services/CO.CDP.OrganisationInformation.Persistence/OrganisationInformationContext.cs
```diff
@@ -2,15 +2,23 @@
...
 {
-    public DbSet<Tenant> Tenants { get; set; } = null!;
-    public DbSet<Organisation> Organisations { get; set; } = null!;
-    public DbSet<Person> Persons { get; set; } = null!;
+    private readonly string _connectionString;
+
+    public OrganisationInformationContext(string connectionString)
+    {
+        _connectionString = connectionString;
+    }
```
jakzal, line 12: Seems more verbose. Any reason for this change?

Services/CO.CDP.Person.WebApi/Api/Person.cs
```diff
@@ -106,6 +107,26 @@ await useCase.Execute(command)
                 operation.Responses["204"].Description = "Person deleted.";
                 return operation;
             });
+
+        app.MapPost("/persons/{personId}/associate",
```
jakzal, line 111: Shouldn't this be part of the Tenant Api? There seems to be a stub for it in there: `"/tenants/{tenantId}/assign-user`

jakzal: The naming there is confusing though (user vs person?) so it might need tweaking.

Services/CO.CDP.Person.WebApi/Api/Person.cs
```diff
@@ -106,6 +107,26 @@ await useCase.Execute(command)
...
+        async (Guid personId, AssociatePerson command, IUseCase<AssociatePerson, Model.Person> useCase) =>
+                 await useCase.Execute(command)
+                     .AndThen(organisation =>
+                         organisation != null
```
jakzal, line 115: According to the `IUseCase<AssociatePerson, Model.Person>` type, person will never be `null`.

Services/CO.CDP.OrganisationInformation.Persistence/OrganisationInformationContext.cs
glgoaco: Ignore this, was part of debugging the issues with EF context

PR #85, opened 2024-05-06
DP-1 Update make file

webit4me:
- Add target to push to ECR (development only for now)
  - Add help
  - Correct some phony target names

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,37 +1,50 @@
-build:
+# Define Docker/ECR attributes
+REPO_URL := 471112892058.dkr.ecr.eu-west-2.amazonaws.com
+IMAGES := cdp-organisation-information-migrations cdp-data-sharing cdp-forms cdp-organisation-app cdp-organisation cdp-person cdp-tenant
+TAGGED_IMAGES := $(addprefix cabinetoffice/,$(addsuffix :latest,$(IMAGES)))
+TAGGED_REPO_IMAGES := $(addprefix $(REPO_URL)/,$(TAGGED_IMAGES))
+
+# Extracts targets and their comments
+help: ## List available commands
+	@echo "Available commands:"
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,37 +1,50 @@
-build:
+# Define Docker/ECR attributes
+REPO_URL := 471112892058.dkr.ecr.eu-west-2.amazonaws.com
+IMAGES := cdp-organisation-information-migrations cdp-data-sharing cdp-forms cdp-organisation-app cdp-organisation cdp-person cdp-tenant
+TAGGED_IMAGES := $(addprefix cabinetoffice/,$(addsuffix :latest,$(IMAGES)))
```
jakzal, line 4: Eventually, we should stop using `latest` version and make each build has its own identifiable version.

jakzal: By the way, compose already tags images (by default with the `latest` tag). It looks for [`IMAGE_VERSION`](https://github.com/cabinetoffice/GCGS-Central-Digital-Platform/blob/426a5c8f9b30980291757e1ae3afea959173481d/compose.common.yml#L20) environment variable.

jakzal (approved):

Makefile
webit4me: Yes, eventually, we will also need to store images in a shared or production account, but we are still far from setting up CI pipelines. This setup is intended to enable me to quickly bring up a demo environment as soon as possible.

webit4me: > By the way, compose already tags images (by default with the `latest` tag). It looks for [`IMAGE_VERSION`](https://github.com/cabinetoffice/GCGS-Central-Digital-Platform/blob/426a5c8f9b30980291757e1ae3afea959173481d/compose.common.yml#L20) environment variable.

Nice! It would be beneficial if the versioning logic mechanism is already in place by the time we start building the CI/CD pipelines. This way, we can avoid spending time on it during that phase.

PR #110, opened 2024-05-14
DP-65 Person is associated with the organisation (entity timestamps)

glgoaco:


Services/CO.CDP.OrganisationInformation.Persistence.Tests/DatabasePersonRepositoryTest.cs
```diff
@@ -71,14 +71,19 @@ public async Task ItUpdatesAnExistingPerson()
...
+        var initialDate = DateTime.UtcNow.AddDays(-1);
+        var person = GivenPerson(guid: guid, email: initialEmail);
+
+        person.Email = "updated@example.com";
         repository.Save(person);
 
         var found = await repository.Find(person.Guid);
 
         found.Should().NotBeNull();
-        found!.Email.Should().Be("email@email.com");
```
jakzal, line 81: To avoid `!` which usually raises a red flag, we could use `found.As<Person>.Email.Should().Be("email@email.com")`.

dharmverma (approved):

PR #125, opened 2024-05-22
DP-131 Consistent error handling in the Web APIs (Organisations POST)

glgoaco:


Libraries/CO.CDP.AspNetCore/ErrorHandlerMiddleware.cs
```diff
@@ -0,0 +1,66 @@
...
+            BadHttpRequestException => StatusCodes.Status422UnprocessableEntity,
+            _ => (int)HttpStatusCode.InternalServerError,
+        };
+
+        var problemDetails = new ProblemDetails
+        {
+            Status = statusCode,
+            Title = statusCode == (int)HttpStatusCode.InternalServerError ? "An unexpected error occurred" : exception.Message,
+            Detail = _environment.IsDevelopment() ? exception.ToString() : (statusCode == (int)HttpStatusCode.InternalServerError ? "Internal Server Error" : exception.Message),
+        };
```
jakzal, line 60: One disadvantage of doing the mapping manually like this instead of using the [built-in mechanisms](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/error-handling?view=aspnetcore-8.0) (`services.AddProblemDetails()` and `app.UseExceptionHandler();`) is that we lose the type:

```
{
  "type": "https://tools.ietf.org/html/rfc9110#section-15.5.1",
  "title": "Bad Request",
  "status": 400
}
```

jakzal: On the other hand, we need to pass the type of error back to the caller, as the frontend will need to map to its own error message at least for some errors. Should we be using custom `type` to pass this information back?

PR #129, opened 2024-05-23
Feature/dp 130 error handling 4xx

JBaigGoaco:
dp 130 error handling 4xx
400 Bad Request
401 Unauthrosed
404 Not found

500 Server Side error was already merged in the previous PR with a redirect page.

### DIFF
--- a/Frontend/CO.CDP.OrganisationApp/ExceptionMiddleware.cs
+++ b/Frontend/CO.CDP.OrganisationApp/ExceptionMiddleware.cs
@@ -1,3 +1,5 @@
+using CO.CDP.Organisation.WebApiClient;
+
 namespace CO.CDP.OrganisationApp;
 
 public class ExceptionMiddleware
@@ -14,9 +16,35 @@ public class ExceptionMiddleware
         {
             await this._next.Invoke(context);
         }
-        catch
+        catch (Exception ex)
         {
-            context.Response.Redirect("/error");
+            await HandleExceptionAsync(context, ex);
+        }
+    }
+    private static Task HandleExceptionAsync(HttpContext context, Exception exception)
+    {
+        context.Response.ContentType = "application/json";
+
+        switch (exception)
+        {
+            case ApiException apiEx when apiEx.StatusCode == 400:
+                context.Response.StatusCode = StatusCodes.Status400BadRequest;
+                return context.Response.WriteAsync(
+                    System.Text.Json.JsonSerializer.Serialize(new { error = "Invalid data provided. Please check your inputs and try again." }));
+
+            case ApiException apiEx when apiEx.StatusCode == 401:
+                context.Response.StatusCode = StatusCodes.Status401Unauthorized;
+                return context.Response.WriteAsync(
+                    System.Text.Json.JsonSerializer.Serialize(new { error = "You are not authorised to perform this action. Please log in and try again." }));
+
+            case ApiException apiEx when apiEx.StatusCode == 404:
+                context.Response.StatusCode = StatusCodes.Status404NotFound;
+                return context.Response.WriteAsync(
+                    System.Text.Json.JsonSerializer.Serialize(new { error = "The requested resource was not found. Please try again." }));
+
+            default:
+                context.Response.Redirect("/error");
+                return Task.CompletedTask;
         }
     }
 }
--- a/Frontend/CO.CDP.OrganisationApp/Helpers/ApiHelper.cs
+++ b/Frontend/CO.CDP.OrganisationApp/Helpers/ApiHelper.cs
@@ -1,0 +1,30 @@
+using CO.CDP.Organisation.WebApiClient;
+
+namespace CO.CDP.OrganisationApp.Helpers;
+
+public static class ApiHelper
+{
+    public static async Task<T> CallApiAsync<T>(Func<Task<T>> apiCall, string errorMessage)
+    {
+        try
+        {
+            return await apiCall();
+        }
+        catch (ApiException apiEx) when (apiEx.StatusCode == 400)
+        {
+            throw new ApiException(errorMessage + " Invalid data provided.", apiEx.StatusCode, apiEx.Response, apiEx.Headers, apiEx);
+        }
+        catch (ApiException apiEx) when (apiEx.StatusCode == 401)
+        {
+            throw new ApiException(errorMessage + " Unauthorized access.", apiEx.StatusCode, apiEx.Response, apiEx.Headers, apiEx);
+        }
+        catch (ApiException apiEx) when (apiEx.StatusCode == 404)
+        {
+            throw new ApiException(errorMessage + " Resource not found.", apiEx.StatusCode, apiEx.Response, apiEx.Headers, apiEx);
+        }
+        catch (Exception ex)
+        {
+            throw new Exception(errorMessage + " An unexpected error occurred.", ex);
+        }
+    }
+}
--- a/Frontend/CO.CDP.OrganisationApp/Pages/Registration/OrganisationDetailsSummary.cshtml.cs
+++ b/Frontend/CO.CDP.OrganisationApp/Pages/Registration/OrganisationDetailsSummary.cshtml.cs
@@ -1,4 +1,5 @@
 using CO.CDP.Organisation.WebApiClient;
+using CO.CDP.OrganisationApp.Helpers;
 using CO.CDP.OrganisationApp.Models;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
@@ -14,7 +15,6 @@ public class OrganisationDetailsSummaryModel(
 {
     public required RegistrationDetails Details { get; set; }
 
-    [BindProperty]
     public string? Error { get; set; }
 
     public void OnGet()
@@ -26,38 +26,39 @@ public class OrganisationDetailsSummaryModel(
     {
         Details = VerifySession();
 
-        var organisation = await RegisterOrganisation(Details);
-        if (organisation != null)
-        {
-            Details.OrganisationId = organisation.Id;
-            session.Set(Session.RegistrationDetailsKey, Details);
-        }
-        else
-        {
-            return Page();
-        }
-
-        return RedirectToPage("OrganisationOverview", new { organisation.Id });
-    }
-
-    private async Task<OrganisationWebApiClient.Organisation?> RegisterOrganisation(RegistrationDetails details)
-    {
         try
         {
-            var payload = NewOrganisationPayload(details);
-            if (payload is not null)
+            var organisation = await ApiHelper.CallApiAsync(
+                        () => RegisterOrganisation(Details),
+                        "Failed to register organisation."
+            );
+
+            if (organisation != null)
             {
-                return await organisationClient.CreateOrganisationAsync(payload);
+                Details.OrganisationId = organisation.Id;
+                session.Set(Session.RegistrationDetailsKey, Details);
             }
-        }
-        catch (ApiException aex)
-            when (aex is ApiException<OrganisationWebApiClient.ProblemDetails> pd)
-        {
-            Error = pd.Result.Detail;
+            else
+            {
+                return Page();
+            }
+
+            return RedirectToPage("OrganisationOverview", new { organisation.Id });
         }
         catch (Exception ex)
         {
             Error = ex.Message;
+            return Page();
+        }
+    }
+
+    private async Task<OrganisationWebApiClient.Organisation?> RegisterOrganisation(RegistrationDetails details)
+    {
+        var payload = NewOrganisationPayload(details);
+
+        if (payload is not null)
+        {
+            return await organisationClient.CreateOrganisationAsync(payload);
         }
 
         return null;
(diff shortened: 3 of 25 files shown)


Frontend/CO.CDP.OrganisationApp/ExceptionMiddleware.cs
```diff
@@ -14,9 +16,35 @@ public async Task Invoke(HttpContext context)
...
         {
-            context.Response.Redirect("/error");
+            await HandleExceptionAsync(context, ex);
+        }
+    }
+    private static Task HandleExceptionAsync(HttpContext context, Exception exception)
+    {
+        context.Response.ContentType = "application/json";
+
+        switch (exception)
```
dharmverma, line 28: I think 4xx errors shouldn't be handled this way, 
e.g. Unauthorised response from API should directly go to error page.
Not found response should be handled from the page itself rather propagated to the general exception middleware.
And certain Bad request (e.g. duplicate org name) should also be handles from the page and allow user to retry.

Frontend/CO.CDP.OrganisationApp/Helpers/ApiHelper.cs
```diff
@@ -0,0 +1,30 @@
+using CO.CDP.Organisation.WebApiClient;
+
+namespace CO.CDP.OrganisationApp.Helpers;
+
+public static class ApiHelper
+{
+    public static async Task<T> CallApiAsync<T>(Func<Task<T>> apiCall, string errorMessage)
```
dharmverma, line 7: This is only handling errors from Organisation API only, are why are we rethrowing them again.

JBaigGoaco: After our discussion, I have removed it now.

Frontend/CO.CDP.OrganisationApp/ExceptionMiddleware.cs
JBaigGoaco: ExceptionMiddleware is now removed and we are handling the error on the page level based on the API.

Frontend/CO.CDP.OrganisationApp/Pages/Registration/OrganisationDetailsSummary.cshtml.cs
```diff
@@ -42,22 +43,29 @@ public async Task<IActionResult> OnPost()
...
-        }
-        catch (ApiException aex)
-            when (aex is ApiException<OrganisationWebApiClient.ProblemDetails> pd)
-        {
-            Error = pd.Result.Detail;
-        }
-        catch (Exception ex)
-        {
-            Error = ex.Message;
+            catch (ApiException<OrganisationWebApiClient.ProblemDetails> aex)
```
dharmverma, line 54: Don't we need to show error page, when there is 500 error from api.
Also code can be shorten too - 

```
if (payload is not null)
{
    try
    {
        return await organisationClient.CreateOrganisationAsync(payload);
    }
    catch (ApiException<OrganisationWebApiClient.ProblemDetails> aex)
        when (aex.StatusCode == StatusCodes.Status400BadRequest)
    {
        ModelState.AddModelError(string.Empty, ErrorMessagesList.OrganisationCreationFailed);
    }
}
```

Services/CO.CDP.Organisation.WebApi/Api/Organisation.cs
```diff
@@ -47,6 +47,7 @@ public static void UseOrganisationEndpoints(this WebApplication app)
                 await useCase.Execute(userUrn)
                     .AndThen(organisations => Results.Ok(organisations)))
             .Produces<List<Model.Organisation>>(200, "application/json")
+            .Produces(404)
```
dharmverma, line 50: List organisation should never return 404

Services/CO.CDP.Organisation.WebApi/Api/Organisation.cs
```diff
@@ -56,14 +57,15 @@ await useCase.Execute(userUrn)
                 return operation;
             });
         app.MapPost("/organisations", async (RegisterOrganisation command, IUseCase<RegisterOrganisation, Model.Organisation> useCase) =>
-            await useCase.Execute(command)
-                .AndThen(organisation =>
-                    organisation != null
-                        ? Results.Created(new Uri($"/organisations/{organisation.Id}", UriKind.Relative), organisation)
-                        : Results.Problem("Organisation could not be created due to an internal error"))
-        )
+        {
```
dharmverma, line 60: Any advantage in this refactor?

Services/CO.CDP.Organisation.WebApi/Api/Organisation.cs
```diff
@@ -111,10 +119,15 @@ await useCase.Execute(organisationId)
...
             {
+                if (!_organisations.ContainsKey(organisationId))
+                {
+                    return Results.NotFound();
+                }
                 _organisations.Remove(organisationId);
                 return Results.NoContent();
             })
             .Produces(204)
+            .Produces(404)
```
dharmverma, line 130: Should 404 be returned from Delete?
https://stackoverflow.com/questions/6439416/status-code-when-deleting-a-resource-using-http-delete-for-the-second-time

Services/CO.CDP.Person.WebApi/Api/Person.cs
```diff
@@ -22,12 +22,21 @@ public static class EndpointExtensions
     public static void UsePersonEndpoints(this WebApplication app)
     {
         app.MapPost("/persons", async (RegisterPerson command, IUseCase<RegisterPerson, Model.Person> useCase) =>
-            await useCase.Execute(command)
-                .AndThen(person =>
-                    Results.Created(new Uri($"/persons/{person.Id}", UriKind.Relative), person))
-                )
+        {
+            if (command == null)
```
dharmverma, line 26: command will never be null here from ASP.NET pipeline

Services/CO.CDP.Person.WebApi/Api/Person.cs
```diff
@@ -78,6 +99,7 @@ await useCase.Execute(personId)
             return Results.NoContent();
         })
             .Produces(204)
+            .Produces(404)
```
dharmverma, line 102: Should 404 be returned from Delete?
https://stackoverflow.com/questions/6439416/status-code-when-deleting-a-resource-using-http-delete-for-the-second-time

Services/CO.CDP.Person.WebApi/Api/Person.cs
```diff
@@ -54,6 +63,16 @@ await useCase.Execute(personId)
         app.MapPut("/persons/{personId}",
                 (Guid personId, UpdatePerson updatedPerson) =>
                 {
+                    if (!_persons.ContainsKey(personId))
+                    {
+                        return Results.NotFound();
+                    }
+
+                    if (updatedPerson == null)
```
dharmverma, line 71: updatedPerson will never be null from ASP.NET pipeline.
Also, BadRequest check should be before 404

Services/CO.CDP.Tenant.WebApi/Api/Tenant.cs
```diff
@@ -66,6 +66,7 @@ public static void UseTenantEndpoints(this WebApplication app)
     {
         app.MapGet("/tenants", () => _tenants.Values.ToArray())
             .Produces<List<Model.Tenant>>(200, "application/json")
+            .Produces(404)
```
dharmverma, line 69: list response should not be 404

Frontend/CO.CDP.OrganisationApp/Pages/Registration/OrganisationSelection.cshtml.cs
```diff
@@ -13,11 +14,31 @@ public class OrganisationSelectionModel(
...
+    public string? Error { get; set; }
+
     public async Task<IActionResult> OnGet()
     {
         var details = VerifySession();
 
-        Organisations = await organisationClient.ListOrganisationsAsync(details.UserUrn);
+        try
+        {
+            Organisations = await organisationClient.ListOrganisationsAsync(details.UserUrn);
```
dharmverma, line 25: List organisation response should never be 404

Frontend/CO.CDP.OrganisationApp/Pages/Registration/OrganisationDetailsSummary.cshtml.cs
JBaigGoaco: Thanks Updated this.

Services/CO.CDP.Organisation.WebApi/Api/Organisation.cs
JBaigGoaco: Thats true 404 will not come from a list. I have removed it.

Services/CO.CDP.Organisation.WebApi/Api/Organisation.cs
JBaigGoaco: I have reverted it back to original.

Services/CO.CDP.Organisation.WebApi/Api/Organisation.cs
JBaigGoaco: I will remove it - in our case it might not the case - it only happens in case user try to delete the same resource twice. but in the UI we should navigate away after the first deletion. I will remove this.

Services/CO.CDP.Person.WebApi/Api/Person.cs
JBaigGoaco: I have removed that command check now.

Services/CO.CDP.Person.WebApi/Api/Person.cs
JBaigGoaco: I guess we will handle the UI and will not let the users delete the same resource twice. I will remove it now.

Services/CO.CDP.Tenant.WebApi/Api/Tenant.cs
JBaigGoaco: true. I have updated it.

Frontend/CO.CDP.OrganisationApp/Pages/Registration/OrganisationSelection.cshtml.cs
JBaigGoaco: true. I have updated this,

Frontend/CO.CDP.OrganisationApp/Pages/Registration/YourDetails.cshtml.cs
```diff
@@ -76,23 +77,37 @@ public async Task<IActionResult> OnPost()
...
+        catch (ApiException<Person.WebApiClient.ProblemDetails> aex)
         {
-            Error = pd.Result.Detail;
+            var statusCode = aex.StatusCode;
+            var problemDetails = aex.Result;
+
+            switch (statusCode)
+            {
+                case StatusCodes.Status400BadRequest:
+                    ModelState.AddModelError(string.Empty, ErrorMessagesList.DuplicatePersonName);
```
jakzal, line 88: What if we add more validation rules to the Person API? 400 will no longer map to a single error. Wouldn't it be better to inspect the actual error type in `ProblemDeatils` and map it to the error message?

Services/CO.CDP.Person.WebApi/Api/Person.cs
```diff
@@ -22,12 +22,21 @@ public static class EndpointExtensions
...
-                .AndThen(person =>
-                    Results.Created(new Uri($"/persons/{person.Id}", UriKind.Relative), person))
-                )
+        {
+            if (command == null)
+            {
+                return Results.BadRequest("Invalid request payload.");
+            }
+
+            var result = await useCase.Execute(command);
```
jakzal, line 31: Can result ever be null?

Frontend/CO.CDP.OrganisationApp/Pages/Registration/YourDetails.cshtml.cs
dharmverma: Than potentially code can be written as 

```
catch (ApiException<Person.WebApiClient.ProblemDetails> aex)
    when (aex.StatusCode == StatusCodes.Status400BadRequest)
{
    var message = aex.Result.Detail switch
    {
        "Err1" => "ui_error_message",
        _ => "default_ui_error_message",
    };

    ModelState.AddModelError(string.Empty, message);
}
```

Services/CO.CDP.Person.WebApi/Api/Person.cs
JBaigGoaco: thanks Kuba, its already been removed.

Frontend/CO.CDP.OrganisationApp/Pages/Registration/YourDetails.cshtml.cs
dharmverma: @JBaigGoaco this comment is resolved but there is still work to be done to map error from API.
Are you creating another story for this?

JBaigGoaco: yes sure. I will add another story on in the sprint backlog with our discussion and comments. to cover the error mapping from the API side of errors. For example for 400 we need to map error codes returned from api, There could be different messages for 400 error.

JBaigGoaco: Story created on the Borad: https://noticingsystem.atlassian.net/browse/DP-168

dharmverma (approved):

dpatel017 (approved):

PR #130, opened 2024-05-23
WebApi Authentication with OneLogin

dharmverma:


Services/CO.CDP.Organisation.WebApi/Api/Organisation.cs
```diff
@@ -174,8 +190,23 @@ public static void DocumentOrganisationApi(this SwaggerGenOptions options)
...
             return typeMap.GetValueOrDefault(type, type.Name);
-        }
-       );
+        });
+
+        var jwtSecurityScheme = new OpenApiSecurityScheme
+        {
+            In = ParameterLocation.Header,
+            Description = "GOV.UK One Login JWT Bearer token",
+            Type = SecuritySchemeType.Http,
```
jakzal, line 199: I used OpenIdConnect with the datasharing stub definition. Was I wrong? https://github.com/cabinetoffice/GCGS-Central-Digital-Platform/pull/128/files#diff-be76f1a108385e00b113e41603b4498a166a2a78f1e48336a75c95a3b288a0dbR243

dharmverma: You are not wrong,  I specified Bearer Token because that what out API was all concerned about and expects it as part of request. OpenIdConnect defines broader definition, of course bearer token is part of that. so specifying that is also not wrong.
Another note that OpenIdConnect discovery endpoint in swagger definition will exposes authority to open world.

jakzal: Ok, I will change mine to Bearer token. It's narrower.

PR #137, opened 2024-05-24
DP-131 Consistent error handling in the Web APIs ProblemDetails

glgoaco:


Libraries/CO.CDP.AspNetCore/CO.CDP.AspNetCore.csproj
```diff
@@ -0,0 +1,15 @@
...
+
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.App" />
+    <ProjectReference Include="..\..\Services\CO.CDP.OrganisationInformation.Persistence\CO.CDP.OrganisationInformation.Persistence.csproj" />
```
jakzal, line 11: If you're planning to make this library dependent on all the projects with exceptions that we plan to map, it's likely to introduce circular dependencies. Remember that exceptions might be thrown by any layer (not only persistence).

Libraries/CO.CDP.AspNetCore/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+
+public static class ServiceCollectionExtensions
+{
+    public static IServiceCollection AddCDPProblemDetails(this IServiceCollection services)
+    {
+        services.AddProblemDetails(options =>
+        {
+            options.CustomizeProblemDetails = ctx =>
+            {
+                switch (ctx.Exception)
```
jakzal, line 19: This is a slippery slope to a God Switch Statement (tm) that knows about every single package in the project.

jakzal: Let me share some red flags I noticed so far.

All the Web API will depend on this component, which makes it `stable`. Web API components are `volatile` as they will change a lot to implement the always changing business landscape.

How would we handle an exception thrown in Web API's use case service?

The Stable-Dependencies Principle (SDP) teaches us:

> Depend in the direction of stability.

Web APIs will depend on CO.CDP.AspNetCore, that's unavoidable. That makes the AspNetCore package the stable one.

As a result CO.CDP.AspNetCore should not depend on Web APIs. With the current design, it will need to.

That will also lead to cyclic dependencies, while The Acyclic Dependencies Principle (ADP) suggests:

> Allow no cycles in the component dependency graph.

Finally, using specific types in this component makes it less abstract, while the Stable-Abstractions Principle (SAP) says:

> A component should be as abstract as it is stable.

I'm quoting principles, as that's their purpose. For a detailed explanation, see [Chapter 28 of Agile Principles, Patterns, and Practices in C#](https://learning.oreilly.com/library/view/agile-principles-patterns/0131857258/ch28.xhtml) or numerous blog posts on Principles of Package and Component Design.

jakzal: What can we do to make this component a bit more abstract? Specifically, make it doesn't depend on specific exception types from other application components?

Services/CO.CDP.OrganisationInformation.Persistence/DatabaseOrganisationRepository.cs
```diff
@@ -58,6 +58,9 @@ private static void HandleDbUpdateException(Organisation organisation, DbUpdateE
             case { } e when e.Message.Contains("_Organisations_Guid"):
                 throw new IOrganisationRepository.OrganisationRepositoryException.DuplicateOrganisationException(
                     $"Organisation with guid `{organisation.Guid}` already exists.", cause);
+            case { } e when e.ContainsDuplicateKey("_Tenants_Name"):
+                throw new ITenantRepository.TenantRepositoryException.DuplicateTenantException(
+                    $"Tenant with name `{organisation.Tenant.Name}` already exists.", cause);
```
jakzal, line 63: Oh.. this **should** not happen in practice 🤔

jakzal: Can we remove this change from this PR and handle it separately?

Services/CO.CDP.Organisation.WebApi.Tests/Extensions/ServiceCollectionExtensionsTests.cs
```diff
@@ -0,0 +1,77 @@
...
+namespace CO.CDP.Organisation.WebApi.Tests.Extensions;
+
+public class ServiceCollectionExtensionsTests
+{
+    private static (int status, string error) InvokeMapException(Exception exception)
+    {
+        var methodInfo = typeof(CO.CDP.Organisation.WebApi.Extensions.ServiceCollectionExtensions)
+            .GetMethod("MapException", BindingFlags.NonPublic | BindingFlags.Static);
+        return ((int status, string error))methodInfo.Invoke(null, new object[] { exception });
+    }
```
jakzal, line 16: There's lots of null related problems reported by static analysis here. 

In general, private methods should be tested through the public interface, not directly. Otherwise, there's no point of them being private and it makes refactoring harder.

Either test this behaviour through the public interface, or make the `MapException` method public please. I think the latter might be actually suitable, as we will need this mapping to later document error responses.

Services/CO.CDP.Organisation.WebApi.Tests/Extensions/ServiceCollectionExtensionsTests.cs
```diff
@@ -0,0 +1,77 @@
...
+
+  [Fact]
+        public void MapException_Should_Return_UnprocessableEntity_For_BadHttpRequestException()
```
jakzal, line 19: Something went wrong with formatting here and below.

jakzal (approved):

PR #143, opened 2024-05-28
Fix organisation registration

jakzal:
re https://github.com/cabinetoffice/GCGS-Central-Digital-Platform/pull/134

I made too many fields non-nullable.


dharmverma (approved):
Are we keeping StreetAddress2 and Region stored as empty in Address DB entity? 
Because that is marked required but at least Region is not captured from UI and StreetAddress2 is optional.

glgoaco (approved):

PR #144, opened 2024-05-29
Allow multiple address for Organisation

dharmverma:
Allow multiple address for Organisation

### DIFF
--- a/Services/CO.CDP.Organisation.WebApi/Model/Command.cs
+++ b/Services/CO.CDP.Organisation.WebApi/Model/Command.cs
@@ -93,7 +93,6 @@ public static class MappingExtensions
             StreetAddress = command.StreetAddress,
             StreetAddress2 = command.StreetAddress2 ?? "",
             Locality = command.Locality,
-            Region = "",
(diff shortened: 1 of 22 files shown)


dpatel017 (approved):

JBaigGoaco (approved):

Services/CO.CDP.Organisation.WebApi/Model/Command.cs
```diff
@@ -93,7 +93,6 @@ public static class MappingExtensions
             StreetAddress = command.StreetAddress,
             StreetAddress2 = command.StreetAddress2 ?? "",
             Locality = command.Locality,
-            Region = "",
```
jakzal, line 96: This might need to eventually return. I saw Region used in the latest prototype.

jakzal (approved):

Services/CO.CDP.Organisation.WebApi/Model/Command.cs
dharmverma: Yes introduced region in organisation API

PR #161, opened 2024-05-30
DP-158 - Basic Information - Individual Or Organisation?

dpatel017:
DP-158 - Basic Information - Individual Or Organisation?


Frontend/CO.CDP.OrganisationApp.Tests/Pages/Registration/OrganisationTypeTest.cs
```diff
@@ -11,11 +11,11 @@
 
 namespace CO.CDP.OrganisationApp.Tests.Pages.Registration;
 
-public class OrganisationTypeTest
+public class SupplierTypeTest
```
jakzal, line 14: This change seems to be an accident?

PR #185, opened 2024-06-06
Buyer & Supplier Database, fron end work

dpatel017:
DP-126
DP-173 - fron end work
DP-158 - front end work
Database migrations for buyer and supplier journey

Renamed OrganisationIdentifier to Identifier
Added DateEntity


dharmverma (approved):

Frontend/CO.CDP.OrganisationApp/Constants/DevolvedRegulation.cs
```diff
@@ -0,0 +1,35 @@
...
+    //public static Organisation.WebApiClient.DevolvedRegulation AsApiDevolvedRegulations(this DevolvedRegulation devolvedRegulations)
+    //{
+    //    return devolvedRegulations switch
+    //    {
+    //        DevolvedRegulation.NorthernIreland => Organisation.WebApiClient.DevolvedRegulation.NorthernIreland,
+    //        DevolvedRegulation.Scotland => Organisation.WebApiClient.DevolvedRegulation.Scotland,
+    //        DevolvedRegulation.Wales => Organisation.WebApiClient.DevolvedRegulation.Wales,
+    //        _ => throw new NotImplementedException(),
+    //    };
+    //}
```
jakzal, line 23: No commented-out code, please!

Services/CO.CDP.Organisation.WebApi/Model/Organisation.cs
```diff
@@ -23,4 +23,6 @@ public record Organisation
 
     /// <example>["Supplier"]</example>
     public required List<PartyRole> Roles { get; init; }
-}
\ No newline at end of file
+    //public string? BuyerType { get; set; }
+    //public List<DevolvedRegulation> DevolvedRegulations { get; set; } = [];
```
jakzal, line 27: Remove commented-out code, please.

Services/CO.CDP.OrganisationInformation.Persistence.Tests/DatabaseOrganisationRepositoryTest.cs
```diff
@@ -153,7 +153,7 @@ public async Task ItUpdatesAnExistingOrganisation()
         updatedOrganisation.Should().NotBeNull();
         updatedOrganisation.As<Organisation>().Name.Should().Be(updatedName);
         updatedOrganisation.As<Organisation>().Tenant.Should().Be(organisation.Tenant);
-        updatedOrganisation.As<Organisation>().UpdatedOn.Should().BeAfter(initialDate);
+        //updatedOrganisation.As<Organisation>().UpdatedOn.Should().BeAfter(initialDate);
```
jakzal, line 156: Has this stopped working, or is it a flaky test?

Services/CO.CDP.OrganisationInformation.Persistence/Organisation.cs
```diff
@@ -46,4 +50,67 @@ public record OrganisationContactPoint
         public string? Telephone;
         public string? Url;
     }
+
+    [Owned]
+    public record SupplierInformation : IEntityDate
+    {
+        public required SupplierType SupplierType { get; set; }
+        public List<OperationType> OperationTypes { get; set; } = [];
+        public bool? CompletedRegAddress { get; set; }
```
jakzal, line 59: Instead of making all these bools nullable, I suggest we make them non-null but set defaults to false.

Anytime we can provide a reasonable non-nullable default, we should use it.

Services/CO.CDP.OrganisationInformation.Persistence/OrganisationInformationContext.cs
```diff
@@ -23,16 +23,51 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
                 a.Property(ai => ai.Scheme);
                 a.Property(ai => ai.LegalName);
                 a.Property(ai => ai.Uri);
+                a.Property(ai => ai.CreatedOn).IsRequired().HasDefaultValueSql("CURRENT_TIMESTAMP");
+                a.Property(ai => ai.UpdatedOn).IsRequired().HasDefaultValueSql("CURRENT_TIMESTAMP");
```
jakzal, line 27: @glgoaco do we need to handle each of these timestamps manually? 

I thought this part takes care of it based on the IDate interface: 

https://github.com/cabinetoffice/GCGS-Central-Digital-Platform/blob/a9743f4545e74c78d307b6d505679a4de005f179/Services/CO.CDP.OrganisationInformation.Persistence/OrganisationInformationContext.cs#L78-L87

Services/CO.CDP.OrganisationInformation.Persistence/OrganisationInformationContext.cs
```diff
@@ -56,13 +91,13 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
 
     public override int SaveChanges()
     {
-        UpdateTimestamps();
+        //UpdateTimestamps();
```
jakzal, line 94: Any reason to comment this out?

Services/CO.CDP.OrganisationInformation.Persistence/OrganisationInformationContext.cs
dpatel017: for Owned type object it need to be configured inside as Organisation has already implemented IEntityDate interface.

Services/CO.CDP.OrganisationInformation.Persistence/Organisation.cs
dpatel017: We need nullable bool for first time else it will be always No answer.

jakzal: It's `false` the first time. We're not storing an answer here, just the fact that the answer was given. 

For example, 

`CompletedVat` == `false` means the supplier has not answered the question yet.
`CompletedVat` == `true` means they answered the question. The existence of vat identifier tells us how they answered.

For these boolean flags, there's only two states (answer given / answer not given). By using `null` as one of the states, we will only have either `null` or `true`. Imo, it's better to use `false` and `true` 🤷

Services/CO.CDP.OrganisationInformation.Persistence/OrganisationInformationContext.cs
jakzal: Ok, the idea behind these timestamps was to implement it in a way we don't have to add too much code manually. It seems to have gone the other direction. It looks like it's a problem for owned entities.

Let's not try to solve it in this PR. I'll make a note to reconsider the implementation separately.

Services/CO.CDP.OrganisationInformation.Persistence/Organisation.cs
dpatel017: Done

JBaigGoaco (approved):

PR #189, opened 2024-06-07
Apply the snake case naming convention to db tables

jakzal:
This will remove the necessity to quote identifiers when making queries.

Unfortunately, the convention is not applied to sequence names due to a bug: https://github.com/efcore/EFCore.NamingConventions/issues/13

It is not a big inconviniece, since sequences are not usually queried directly.

image

image

image


glgoaco (approved):

dharmverma (approved):

Services/CO.CDP.OrganisationInformation.Persistence/OrganisationInformationContext.cs
```diff
@@ -160,4 +177,21 @@ ValueComparer<TProperty> valueComparer
...
             valueComparer
         );
-}
\ No newline at end of file
+}
+
+/// <a href="https://github.com/efcore/EFCore.NamingConventions/issues/1">Keep history columns CamelCased</a>
+// ReSharper disable once ClassNeverInstantiated.Global
+#pragma warning disable EF1001
+internal class CamelCaseHistoryContext(HistoryRepositoryDependencies dependencies) : NpgsqlHistoryRepository(dependencies)
```
jakzal, line 185: Let's remove this class at the time we squash migrations.

PR #194, opened 2024-06-10
Registration - Buyer Route - Buyer Submission Page

dpatel017:
DP-173 - Registration - Buyer Route - Buyer Submission Page

### DIFF
--- a/Libraries/CO.CDP.Organisation.WebApiClient.Tests/OrganisationClientIntegrationTest.cs
+++ b/Libraries/CO.CDP.Organisation.WebApiClient.Tests/OrganisationClientIntegrationTest.cs
@@ -48,10 +48,15 @@ public class OrganisationClientIntegrationTest(ITestOutputHelper testOutputHelpe
             )
         };
         var roles = new List<PartyRole> { PartyRole.Supplier };
+        var buyerInfo = new BuyerInformation(
+            buyerType:"Buyer Type",
+            devolvedRegulations: [DevolvedRegulation.NorthernIreland, DevolvedRegulation.Wales]);
+
         var newOrganisation = new NewOrganisation(
             personId: unknownPersonId,
             additionalIdentifiers: additionalIdentifiers,
             addresses: [address],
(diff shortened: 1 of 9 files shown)


dharmverma (approved):

JBaigGoaco (approved):

Services/CO.CDP.OrganisationInformation/BuyerInformation.cs
```diff
@@ -0,0 +1,10 @@
+namespace CO.CDP.OrganisationInformation;
+
+public record BuyerInformation
+{
+
+    public string? BuyerType { get; set; }
+
+    public List<DevolvedRegulation> DevolvedRegulations { get; set; } = [];
+    
+}
```
jakzal, line 10: This type isn't meant to be shared between APIs, and should go to the Organisation Web API. 

To avoid polluting this common namespace, it's best to start with types kept in the Web API they originated and only move to CO.CDP.OrganisationInformation through refactoring when they're needed to be shared.

Libraries/CO.CDP.Organisation.WebApiClient.Tests/OrganisationClientIntegrationTest.cs
```diff
@@ -48,10 +48,15 @@ public async Task ItTalksToTheOrganisationApi()
...
         var roles = new List<PartyRole> { PartyRole.Supplier };
+        var buyerInfo = new BuyerInformation(
+            buyerType:"Buyer Type",
+            devolvedRegulations: [DevolvedRegulation.NorthernIreland, DevolvedRegulation.Wales]);
+
         var newOrganisation = new NewOrganisation(
             personId: unknownPersonId,
             additionalIdentifiers: additionalIdentifiers,
             addresses: [address],
+            buyerInfo: buyerInfo,
```
jakzal, line 59: We agreed to have a separate call to `PATCH /organisations/123/buyer-information` to keep buyer and supplier stuff separate from the core organisation details. We should not submit buyer info with the register organisation call.

PR #200, opened 2024-06-12
Supplier Information - Organisation

dharmverma:
Supplier Information - Organisation

### DIFF
--- a/Services/CO.CDP.Organisation.WebApi/Properties/launchSettings.json
+++ b/Services/CO.CDP.Organisation.WebApi/Properties/launchSettings.json
@@ -14,7 +14,7 @@
       "dotnetRunMessages": true,
       "launchBrowser": true,
       "launchUrl": "swagger",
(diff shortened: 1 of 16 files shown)


Services/CO.CDP.Organisation.WebApi/Properties/launchSettings.json
```diff
@@ -14,7 +14,7 @@
       "dotnetRunMessages": true,
       "launchBrowser": true,
       "launchUrl": "swagger",
-      "applicationUrl": "http://localhost:8082",
+      "applicationUrl": "http://localhost:7208",
```
jakzal, line 17: Any reason for this change? Some time ago I aligned these ports with docker. Isn't that what we want?

dharmverma: Make sense, will revert change

PR #210, opened 2024-06-13
DP-168 Tenant Lookup tests

glgoaco:
Additional tests for the scenarios:

- A person with a tenant and organisations 
- A person associated with multiple tenants and organisations 
- A person with no tenant connection 
- A person with a tenant but no organisation connection

### DIFF
--- a/Services/CO.CDP.OrganisationInformation.Persistence/TenantLookup.cs
+++ b/Services/CO.CDP.OrganisationInformation.Persistence/TenantLookup.cs
@@ -25,6 +25,6 @@ public class TenantLookup
         public required Guid Id { get; init; }
         public required string Name { get; init; }
         public required List<PartyRole> Roles { get; init; }
-        public required List<string> Scopes { get; init; }
+        public required List<string>? Scopes { get; init; }
     }
 }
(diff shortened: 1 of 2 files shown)


JBaigGoaco (approved):

Services/CO.CDP.OrganisationInformation.Persistence/TenantLookup.cs
```diff
@@ -25,6 +25,6 @@ public class Organisation
         public required Guid Id { get; init; }
         public required string Name { get; init; }
         public required List<PartyRole> Roles { get; init; }
-        public required List<string> Scopes { get; init; }
+        public required List<string>? Scopes { get; init; }
```
jakzal, line 28: Is it possible to use an empty list rather than null?

Services/CO.CDP.OrganisationInformation.Persistence/DatabaseTenantRepository.cs
```diff
@@ -48,7 +48,8 @@ public void Save(Tenant tenant)
                         Id = o.Guid,
                         Name = o.Name,
                         Roles = o.Roles,
-                        Scopes = o.OrganisationPersons.Single(op => op.PersonId == p.Id).Scopes
+                        // ReSharper disable once NullCoalescingConditionIsAlwaysNotNullAccordingToAPIContract
```
jakzal, line 51: Ignored a rider inspection as it is a false positive. The test proves null value is possible here.

jakzal (approved):

PR #222, opened 2024-06-17
DP-217 Add Grafana image configuration

webit4me:


terragrunt/tools/grafana/README.md
```diff
@@ -0,0 +1,39 @@
...
+#   - AWS_PROFILE
+#   - TG_ENVIRONMENT
+#   - AWS_ENV (Not compulsory)
+#   - MFA_TOKEN handler (Out of scope of this documentation)
+# Feel free to use any convenient AWS profiler instead.
+```
+## build
+
+```shell
+docker build -t cdp-grafana .
```
jakzal, line 18: For consistency, can we use `cabinetoffice/cdp-grafana` tag pretty please?

terragrunt/tools/grafana/configs/entrypoint.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+set -e
+
+sed -i "s/GRAFANA_CLOUDWATCH_ACCOUNT_ID/${GRAFANA_CLOUDWATCH_ACCOUNT_ID}/g" /etc/grafana/provisioning/dashboards/application/*
```
jakzal, line 4: I assume it isn't possible to use env vars in config files?

webit4me: Correct assumption, Grafana's dashboards are plain JSON. so here a simple `sed` does the rendering of the env_var into them

terragrunt/tools/grafana/README.md
webit4me: Like the consistency but I was not too keen about mixing an of-the-shelf tool with our services :thinking: 
Talking about consistency choosing a unified format for our env_vars makes me a very happy person :D 
image

jakzal: That's true. That's because these are just app config options and it's possible to override them by constructing env var name to correspond the config option.

What would you prefer them being named? We could give them a prefix by putting all app settings under a single key (i.e. CDP).

Also, note that these are case insensitive, so you could CAPITALISE them.

dharmverma (approved):

terragrunt/tools/grafana/README.md
webit4me: hmm, being able to capitalise them all makes the maintenance of the infra side a touch easier. 
what about underscores? can we agree on naming like PERSON_SERVICE, instead of PERSONSERVICE.
also what about the number of the underscores? is there a rule like if it is .NET related (Framework Config) they must be double and if us defining (APP Configs) them they must be single?
Perhaps something for team chat instead of here but the current convention make my eyes bleed :D

jakzal: >   what about the number of the underscores?

That is something we cannot change. Double underscore maps to a nested configuration option in .net.

jakzal: > but the current convention make my eyes bleed 

There's no convention. It's accidental.

jakzal (approved):

webit4me:
Closed this in favour of [new PR](https://github.com/cabinetoffice/GCGS-Central-Digital-Platform/pull/237) which address the feedback on the local docker tag under its [last commit ](https://github.com/cabinetoffice/GCGS-Central-Digital-Platform/pull/237/commits/8e1ea8e34c4f8bf87dc85258096215005c690d45) which also adds the Authority dashboard

PR #229, opened 2024-06-20
Configure KnownNetworks

jakzal:
The `FORWARDEDHEADERS__KNOWNNETWORK` environment variable will need to be set to something like `10.3.0.0/16` on AWS.

Known networks won't be re-configured unless the environment variable is set. x-forwarded-for and x-forwarded-proto headers are always added to the supported headers list.

In order to have consistency between projects, I have extracted the logic to a library. It's called `CO.CDP.Configuration` and can be further extended with shared configuration logic.


Services/CO.CDP.Tenant.WebApi/appsettings.json
```diff
@@ -9,5 +9,8 @@
   "Organisation:Authority": "",
   "ConnectionStrings": {
     "OrganisationInformationDatabase": "Server=localhost;Database=cdp;Username=cdp_user;Password=cdp123;Trusted_Connection=True;"
```
glgoaco, line 11: happy to commit this?

glgoaco (approved):

dharmverma (approved):

dpatel017 (approved):

PR #240, opened 2024-06-26
Introduce a local gateway to all services run in docker compose

jakzal:
This way we mirror what is on production and we can log additional details useful in development, like authorization headers.

By default, docker services are configured to connect via the gateway rather than talk to each other directly.

Services that we choose to start with the IDE instead of Docker, are also proxied through the gateway. That's why I had to give them unique port numbers again. This way, we can finally mix docker-started services with IDE-started services however we please.

image


dpatel017 (approved):

JBaigGoaco (approved):

glgoaco (approved):
Tested on Windows 11

dharmverma (approved):

PR #242, opened 2024-06-26
Feature/dp 221 organisation lookup by identifier

JBaigGoaco:
Feature/dp 221 organisation lookup by identifier


glgoaco (approved):

Services/CO.CDP.Organisation.WebApi/Api/Organisation.cs
```diff
@@ -158,25 +158,25 @@ public static RouteGroupBuilder UseOrganisationLookupEndpoints(this RouteGroupBu
...
-                operation.Description = "Find an organisation.";
-                operation.Summary = "Find an organisation.";
-                operation.Tags = new List<OpenApiTag> { new() { Name = "Organisation - Lookup" } };
-                operation.Responses["200"].Description = "Organisations Associated.";
-                operation.Responses["401"].Description = "Valid authentication credentials are missing in the request.";
-                operation.Responses["404"].Description = "Organisation not found.";
-                operation.Responses["500"].Description = "Internal server error.";
-                return operation;
-            });
+           async ([FromQuery] string query, IUseCase<string, Model.Organisation?> useCase) =>
```
jakzal, line 161: We were suppose to add a new query parameter to support both `name` and `identifier`.
It only works here as we can distinguish between the name and identifier, but what if we have to support another query type in a week?

It's actually specified in the ticket to support this kind of query in addition to name lookup: `/organisation/lookup?identifier=GB-COH:9543428757`

JBaigGoaco: I will update it now

JBaigGoaco: All updated now.

Services/CO.CDP.Organisation.WebApi/Api/Organisation.cs
```diff
@@ -157,26 +157,43 @@ public static RouteGroupBuilder UseOrganisationLookupEndpoints(this RouteGroupBu
...
-                operation.Responses["200"].Description = "Organisations Associated.";
-                operation.Responses["401"].Description = "Valid authentication credentials are missing in the request.";
-                operation.Responses["404"].Description = "Organisation not found.";
-                operation.Responses["500"].Description = "Internal server error.";
-                return operation;
-            });
+                if (!string.IsNullOrEmpty(identifier))
+                {
+                    return await useCase.Execute($"identifier:{identifier}")
+                        .AndThen(organisation => organisation != null ? Results.Ok(organisation) : Results.NotFound());
```
jakzal, line 167: Rather than rely on a convention of the string, can we introduce a query object instead, please? 

You can create it in this controller, i.e.:

```csharp
useCase.Execute(new OrganisationQuery(name: name, identifier: identifier)
```

This way, the condition and validation will move to the use case. You'll need to throw an exception if both name and identifier are missing.

JBaigGoaco: @jakzal I have updated your suggestion in the code.

Services/CO.CDP.Organisation.WebApi/UseCase/LookupOrganisationUseCase.cs
```diff
@@ -3,11 +3,58 @@
...
+    public async Task<Model.Organisation?> Execute(OrganisationQuery query)
     {
-        return await organisationRepository.FindByName(name)
-            .AndThen(mapper.Map<Model.Organisation>);
+        if (!query.IsValid)
+        {
+            throw new ArgumentException("Organisation name or identifier must be provided.");
+        }
+
+        if (!string.IsNullOrEmpty(query.Identifier))
```
jakzal, line 15: I think you're missing the test(s) to cover validation. This case is probably already covered by the previous check to `if (!query.IsValid)`.

Services/CO.CDP.Organisation.WebApi/UseCase/LookupOrganisationUseCase.cs
```diff
@@ -3,11 +3,58 @@
...
+public class LookupOrganisationUseCase(IOrganisationRepository organisationRepository, IMapper mapper) : IUseCase<OrganisationQuery, Model.Organisation?>
 {
-    public async Task<Model.Organisation?> Execute(string name)
+    public async Task<Model.Organisation?> Execute(OrganisationQuery query)
     {
-        return await organisationRepository.FindByName(name)
-            .AndThen(mapper.Map<Model.Organisation>);
+        if (!query.IsValid)
+        {
+            throw new ArgumentException("Organisation name or identifier must be provided.");
```
jakzal, line 12: This has to be a more specific (dedicated) exception, so it could be mapped to a 400 error response.

Throwing System exceptions directly, without extending them, should be avoided. The reason for this is such an exception can be thrown by anyone, including third party libraries, and we can't unambiguously map it to a 400 error response.

Services/CO.CDP.Organisation.WebApi/UseCase/LookupOrganisationUseCase.cs
jakzal: Thinking about it more, I might know why you've done it this way. Type checker will complain otherwise. However, this makes `IsValid` check redundant.

Wouldn't it be better if we asked for a valid identifier and a valid name?

For example:

```csharp
public async Task<Model.Organisation?> Execute(OrganisationQuery query)
{
    if (query.TryGetIdentifier(out var scheme, out var id))
    {
          return await organisationRepository.FindByIdentifier(scheme, id)
                                 .AndThen(mapper.Map<Model.Organisation>);
    }

    if (query.TryGetName(out var name))
    {
          return await organisationRepository.FindByName(query.Name)
                                 .AndThen(mapper.Map<Model.Organisation>);
    }

    throw new InvalidQueryException("Both name and identifier are missing from the request.");
}
```

Services/CO.CDP.Organisation.WebApi/UseCase/LookupOrganisationUseCase.cs
```diff
@@ -3,11 +3,58 @@
...
+            id = parts[1];
+            return true;
+        }
+        scheme = string.Empty;
+        id = string.Empty;
+        return false;
+    }
+}
+
+public class OrganisationQuery
```
jakzal, line 48: As this is a value object, a record would be a better choice.

Services/CO.CDP.OrganisationInformation.Persistence/DatabaseOrganisationRepository.cs
```diff
@@ -34,6 +34,14 @@ public async Task<IEnumerable<Organisation>> FindByUserUrn(string userUrn)
...
         return person?.Organisations ?? [];
     }
+    public async Task<Organisation?> FindByIdentifier(string scheme, string identifierId)
+    {
+        return await context.Organisations
+            .Include(p => p.Identifiers)
+            .Include(p => p.Addresses)
+            .ThenInclude(p => p.Address)
+            .FirstOrDefaultAsync(o => o.Identifiers.Any(i => i.Scheme == scheme && i.IdentifierId == identifierId));
+    }
```
jakzal, line 44: I didn't see a test case for this ;)

Services/CO.CDP.Organisation.WebApi/UseCase/LookupOrganisationUseCase.cs
```diff
@@ -3,11 +3,58 @@
 using CO.CDP.OrganisationInformation.Persistence;
 
 namespace CO.CDP.Organisation.WebApi.UseCase;
-public class LookupOrganisationUseCase(IOrganisationRepository organisationRepository, IMapper mapper) : IUseCase<string, Model.Organisation?>
+public class LookupOrganisationUseCase(IOrganisationRepository organisationRepository, IMapper mapper) : IUseCase<OrganisationQuery, Model.Organisation?>
```
jakzal, line 6: Service registration in `Services/CO.CDP.Organisation.WebApi/Program.cs` will need to be updated too.

Services/CO.CDP.Organisation.WebApi/UseCase/LookupOrganisationUseCase.cs
JBaigGoaco: Hi Kuba, Updated this.

Services/CO.CDP.OrganisationInformation.Persistence/DatabaseOrganisationRepository.cs
JBaigGoaco: Hi Kuba, Updated this.

Services/CO.CDP.Organisation.WebApi/UseCase/LookupOrganisationUseCase.cs
JBaigGoaco: Hi Kuba, Updated this.

Services/CO.CDP.Organisation.WebApi/UseCase/LookupOrganisationUseCase.cs
JBaigGoaco: Hi Kuba, Updated this.

Services/CO.CDP.Organisation.WebApi/UseCase/LookupOrganisationUseCase.cs
JBaigGoaco: Hi Kuba, Updated this.

jakzal:
One last thing is to make sure 400 is returned when both name and identifier query parameters are missing.

jakzal (approved):

PR #261, opened 2024-07-07
#dp-234 tenant lookup read urn from JWT token

JBaigGoaco:
#dp-234 tenant lookup read urn from JWT token


Services/CO.CDP.Tenant.WebApi/UseCase/LookupTenantUseCase.cs
```diff
@@ -1,14 +1,23 @@
...
+    : IUseCase<Model.TenantLookup?>
 {
-    public async Task<Model.TenantLookup?> Execute(string userUrn)
+    public async Task<Model.TenantLookup?> Execute()
     {
+        var userUrn = claimService.GetUserUrn();
+
+        if (string.IsNullOrEmpty(userUrn))
+        {
+            throw new UnknownTokenException("Cannot find sub or urn from JWT token.");
```
glgoaco, line 18: Maybe the error could be something like _Ensure the token is valid and contains the necessary claims_

glgoaco (approved):

GCGS-Central-Digital-Platform.slnLaunch
```diff
@@ -0,0 +1,21 @@
...
+        "Action": "Start",
+        "DebugTarget": "http-for-docker"
+      },
+      {
+        "Name": "Libraries\\CO.CDP.Authentication\\CO.CDP.Authentication.csproj",
+        "Action": "Start"
+      }
+    ]
+  }
+]
```
jakzal, line 21: Since this profile only starts a subset of services, seems like it's made for one specific use case. Does it deserve to be comitted to the repo?

Libraries/CO.CDP.Authentication.Tests/ClaimServiceTests.cs
```diff
@@ -0,0 +1,41 @@
...
+{
+    [Fact]
+    public void GetUserUrn_ShouldReturnUrn_WhenUserHasSubClaim()
+    {
+        var userUrn = "urn:fdc:gov.uk:2022:rynbwxUssDAcmU38U5gxd7dBfu9N7KFP9_nqDuZ66Hg";
+        var claims = new List<Claim> { new Claim("sub", userUrn) };
+        var identity = new ClaimsIdentity(claims, "TestAuthType");
+        var claimsPrincipal = new ClaimsPrincipal(identity);
+        var httpContext = new DefaultHttpContext { User = claimsPrincipal };
+        var httpContextAccessor = new Mock<IHttpContextAccessor>();
```
jakzal, line 17: This is a lot of setup code, that hinders the readability of the test. 

All that happens here, if I understood it well, is:

```csharp
var httpContextAccessor = GivenHttpContextWith(userUrn: userUrn)
```

Services/CO.CDP.Tenant.WebApi/Api/Tenant.cs
```diff
@@ -80,6 +81,7 @@ await useCase.Execute(urn)
                 operation.Responses["200"].Description = "Tenants associated with the the user.";
                 operation.Responses["401"].Description = "Valid authentication credentials are missing in the request.";
                 operation.Responses["404"].Description = "Tenant not found.";
+                operation.Responses["417"].Description = "Token sub is not found";
```
jakzal, line 84: This isn't the correct status code to return here. Don't we have 401 for this scenario?

[417](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/417) is for when the expectation given in the `Expect` header isn't met:

> The HTTP 417 Expectation Failed client error response code indicates that the expectation given in the request's [Expect](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Expect) header could not be met.

Services/CO.CDP.Tenant.WebApi/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -10,7 +11,9 @@ public static class ServiceCollectionExtensions
         { typeof(DuplicateTenantException), (StatusCodes.Status400BadRequest, "TENANT_ALREADY_EXISTS") },
         { typeof(ArgumentNullException), (StatusCodes.Status400BadRequest, "ARGUMENT_NULL") },
         { typeof(InvalidOperationException), (StatusCodes.Status400BadRequest, "INVALID_OPERATION") },
-        { typeof(TenantNotFoundException), (StatusCodes.Status404NotFound, "TENANT_DOES_NOT_EXIST") }
+        { typeof(TenantNotFoundException), (StatusCodes.Status404NotFound, "TENANT_DOES_NOT_EXIST") },
+        { typeof(UnknownTokenException), (StatusCodes.Status417ExpectationFailed, "TOKEN_SUB_DOES_NOT_EXIST") },
```
jakzal, line 15: I think it should be 401 and NOT_AUTHENTICATED.

Services/CO.CDP.Tenant.WebApi/UseCase/LookupTenantUseCase.cs
jakzal: The exception name is misleading. What does it mean the token is unknown? In this scenario, the token might or might not be there, but the information we need is missing (urn).

Wouldn't `MissingUserUrnException` better reflect what happened here?

Also, have you considered the ClaimService to throw this exception instead of returning `null`? It would only work if it's never optional though and I'm not so sure about it.

JBaigGoaco: Thanks @glgoaco i will update that.

Services/CO.CDP.Tenant.WebApi/UseCase/IUseCase.cs
```diff
@@ -3,4 +3,9 @@ namespace CO.CDP.Tenant.WebApi.UseCase;
 public interface IUseCase<in TCommand, TOutcome>
 {
     public Task<TOutcome> Execute(TCommand command);
+}
+
+public interface IUseCase<TOutcome>
```
jakzal, line 8: 👌

jakzal (approved):

PR #265, opened 2024-07-08
Feature/dp 221 organisation lookup both name and identifier

JBaigGoaco:
image

### DIFF
--- a/Services/CO.CDP.Organisation.WebApi.Tests/Extensions/ServiceCollectionExtensionsTests.cs
+++ b/Services/CO.CDP.Organisation.WebApi.Tests/Extensions/ServiceCollectionExtensionsTests.cs
@@ -78,7 +78,7 @@ public class ServiceCollectionExtensionsTests
         var result = ServiceCollectionExtensions.MapException(exception);
 
         Assert.Equal(StatusCodes.Status400BadRequest, result.status);
-        Assert.Equal("MISSING_QUERY_PARAMETERS", result.error);
+        Assert.Equal("ISSUE_WITH_QUERY_PARAMETERS", result.error);
     }
 
     [Fact]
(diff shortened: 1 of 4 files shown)


Services/CO.CDP.Organisation.WebApi.Tests/Extensions/ServiceCollectionExtensionsTests.cs
```diff
@@ -78,7 +78,7 @@ public void MapException_Should_Return_BadRequest_For_InvalidQueryException()
         var result = ServiceCollectionExtensions.MapException(exception);
 
         Assert.Equal(StatusCodes.Status400BadRequest, result.status);
-        Assert.Equal("MISSING_QUERY_PARAMETERS", result.error);
+        Assert.Equal("ISSUE_WITH_QUERY_PARAMETERS", result.error);
```
jakzal, line 81: Why changing this test instead of adding a new one? 🤔

jakzal: I see, you changed the code for the existing case.

Services/CO.CDP.Organisation.WebApi/UseCase/LookupOrganisationUseCase.cs
```diff
@@ -9,6 +9,11 @@ public class LookupOrganisationUseCase(IOrganisationRepository organisationRepos
 {
     public async Task<Model.Organisation?> Execute(OrganisationQuery query)
     {
+        if (!string.IsNullOrEmpty(query.Name) && !string.IsNullOrEmpty(query.Identifier))
+        {
+            throw new InvalidQueryException("Both name and identifier cannot be provided together.");
+        }
```
jakzal, line 15: Can we add a test case for this please.

JBaigGoaco: testcase is added

jakzal (approved):

PR #305, opened 2024-07-18
Feature/dp 384 ppon service publishes a ppon generated event

dbgoaco:


glgoaco (approved):

Services/CO.CDP.EntityVerification/Ppon/OrganisationRegisteredEventHandler.cs
```diff
@@ -12,7 +15,14 @@ public Task Handle(OrganisationRegistered @event)
 
         Persistence.Ppon newIdentifier = new() { PponId = pponId };
 
-        pponRepository.Save(newIdentifier);
+        using (TransactionScope txn = new TransactionScope())
+        {
+            pponRepository.Save(newIdentifier);
+            publisher.Publish(newIdentifier);
+
+            txn.Complete();
```
jakzal, line 23: Consider the following scenarios:

| Scenario  | In transaction | `r.Save(i)` | `p.Publish(i)` | Identifier saved | Message Published | Comment |
| ------------- | ------------- |------------- |------------- |------------- |------------- |------------- |
| Happy path (transaction)  | Yes  | 🟢   | 🟢  | 🟢  | 🟢  | |
| Happy path  | No   | 🟢   | 🟢  | 🟢  | 🟢  | |
| Repository fails (transaction)  | Yes  | 🔴 | 🟢 | 🔴 | 🟢 | Repository failure isn't triggered until the transaction is committed. The message is published even though the identifier isn't saved. |
| Repository fails | No  | 🔴  | 🟢  | 🔴  | 🔴  | |
| Publisher fails (transaction)  | Yes  | 🟢 | 🔴 |  🔴 | 🔴 | |
| Publisher fails | No  | 🟢  | 🔴  | 🟢  | 🔴  |  Identifier is saved, but the message isn't published. |


Is it a good idea to use a db transaction while the publisher isn't using the database?

It's all up to differences between "Repository fails (transaction)" and "Publisher fails", as all the other scenarios are as expected. 

Is it better to have a message published for identifier that wasn't saved, or to have an identifier saved and no message published?

As part of improving resiliency (separate story), we're planning to introduce the outbox pattern. That would make the repository and publisher can share the transaction. At this point in time, I think the transaction brings more problems.

jakzal (changes requested):

Services/CO.CDP.EntityVerification/appsettings.json
```diff
@@ -10,13 +10,18 @@
...
   },
   "InboundQueue": {
     "Name": "",
     "MaxNumberOfMessages": 10,
     "WaitTimeSeconds": 20
   },
+  "OutboundQueue": {
+    "Name": "",
+    "MaxNumberOfMessages": 10,
+    "WaitTimeSeconds": 20
```
jakzal, line 23: Do we use these parameters with the Outbound Queue?

Services/CO.CDP.EntityVerification/Ppon/OrganisationRegisteredEventHandler.cs
dbgoaco: It's a shame we're not using a queue tech that supports transactions as this would solve the resiliency problem. I added the TransactionScope as I consider it best practice when going across multiple resources. If we could switch to ActiveMQ which does support transactions then we wouldn't need to implement the outbox pattern but I appreciate the spec says SQS.

Services/CO.CDP.EntityVerification/appsettings.json
dbgoaco: Good point, we don't. I have removed them.

Services/CO.CDP.EntityVerification/Ppon/OrganisationRegisteredEventHandler.cs
dbgoaco: Let's remove the TransactionScope for now and we either re-add if we use e.g. ActiveMQ or we implement the outbox pattern.

compose.yml
```diff
@@ -244,3 +244,7 @@ services:
       ASPNETCORE_ENVIRONMENT: Production
       ConnectionStrings__EvDatabase: 'Server=db;Database=cdp;Username=cdp_user;Password=cdp123;'
       Sqs__ServiceURL: "http://localstack:4566"
+      InboundQueue__Name: "ev-inbound"
+      InboundQueue__MaxNumberOfMessages: 20
+      InboundQueue__WaitTimeSeconds: 20
```
jakzal, line 249: We only override things that are different on Docker environment to limit amount of config to maintain. 

`InboundQueue__MaxNumberOfMessages` and `InboundQueue__WaitTimeSeconds` aren't necessary here. 

Acually, I mislead you before . Even `InboundQueue__Name` isn't needed here. Sorry! Please remove all three.

jakzal: All four, there's the outbond one below.

dbgoaco: Ok they've been removed.

jakzal (approved):

PR #318, opened 2024-07-19
Feature/dp 384 add identifiers

dbgoaco:


rmohammed-goaco (approved):
Looks good to me, but as I'm new too prob should get another review also. I did notice tests failed, but you just pushed an update so hopefully that makes them all pass 👍

Services/CO.CDP.EntityVerification.Tests/Events/EventDeserializerTest.cs
```diff
@@ -10,6 +10,18 @@ public class EventDeserializerTest
     [Fact]
     public void ItDeserializesOrganisationRegisteredEvent()
     {
+        var organisationRegistered1 = new OrganisationRegistered()
+        {
+            Id = Guid.NewGuid(),
+            Name = "name",
+            Identifier = new Identifier() { Id = "1234", LegalName = "name", Scheme = "test" },
```
jakzal, line 17: Can we use examples more meaningful to the reader, please? I.e LegalName = "Acme" Scheme "GB-COH".

Services/CO.CDP.EntityVerification.Tests/Persistence/DatabasePponRepositoryTest.cs
```diff
@@ -27,20 +27,24 @@ public async Task ItFindsSavedPpon()
...
         repository.Save(ppon);
 
         repository.Invoking(r => r.Save(anotherPpon))
             .Should().Throw<DuplicatePponException>();
     }
 
-    private static EntityVerification.Persistence.Ppon GivenPpon(string? pponId = null)
+    private static EntityVerification.Persistence.Ppon GivenPpon(string orgName, string? pponId = null)
     {
+        var ppon = GivenPpon("Acme Corp Ltd", pponId: "d73c9fef3c8549e2b3999fa9ec0a5aea");
```
jakzal, line 41: Is this method calling itself? 🤔

Services/CO.CDP.EntityVerification.Tests/Persistence/DatabasePponRepositoryTest.cs
```diff
@@ -27,20 +27,24 @@ public async Task ItFindsSavedPpon()
...
+        var newIdentifier = new Identifier { Ppon = ppon, LegalName = orgName, Scheme = "" };
```
jakzal, line 42: Something went wrong there. The ppon is its own entity. Identifier is an entity to represent all the other identifiers.

Services/CO.CDP.EntityVerification.Tests/Ppon/OrganisationRegisteredEventHandlerTests.cs
```diff
@@ -13,20 +13,17 @@ public async Task Action_GeneratesPponIdAndPersistsAndPublishes()
...
-        var handler = new OrganisationRegisteredEventHandler(pponService.Object, pponRepository.Object, publisher.Object);
+        var handler = new OrganisationRegisteredEventHandler(pponService.Object, pponRepository.Object);
         var @event = GivenOrganisationRegisteredEvent();
 
         await handler.Handle(@event);
 
         pponRepository.Verify(
             s => s.Save(It.Is<EntityVerification.Persistence.Ppon>(p => p.PponId == generatedPpon)), Times.Once);
-        publisher.Verify(
-              s => s.Publish(It.Is<EntityVerification.Persistence.Ppon>(p => p.PponId == generatedPpon)), Times.Once);
```
jakzal, line 29: Why are we removing this?

dbgoaco: Because that's part of DP-385

Services/CO.CDP.EntityVerification/Ppon/OrganisationRegisteredEventHandler.cs
```diff
@@ -1,21 +1,33 @@
...
+        foreach (var e in @event.AllIdentifiers())
+        {
+            ids.Add(new Identifier { Ppon = newPpon,
+                LegalName = e.LegalName,
+                Scheme = e.Scheme,
+                Uri = e.Uri });
+        }
 
-        pponRepository.Save(newIdentifier);
-        await publisher.Publish(newIdentifier);
```
jakzal, line 19: I just realised, we need a proper event type to publish that is separate from the Ppon model. We must not be publishing EF entities!

Services/CO.CDP.EntityVerification.Tests/Events/EventDeserializerTest.cs
```diff
@@ -10,6 +10,18 @@ public class EventDeserializerTest
...
+            Roles = new List<string>(),
+            AdditionalIdentifiers = new List<Identifier>()
+        };
+
+        string s = JsonSerializer.Serialize(organisationRegistered1);
```
jakzal, line 22: Looks like this isn't asserted on anywhere. Why create a new value? just alter the one below `organisationRegistered`.

Services/CO.CDP.EntityVerification.Tests/Events/EventDeserializerTest.cs
dbgoaco: Ah that was a mistake (can't think in this heat), it was there so I could get the json to do some end to end testing with local stack. Will remove it.

Services/CO.CDP.EntityVerification.Tests/Ppon/OrganisationRegisteredEventHandlerTests.cs
jakzal: Sure, but we shouldn't be removing previously added features :)

Services/CO.CDP.EntityVerification/CO.CDP.EntityVerification.csproj
```diff
@@ -22,12 +22,13 @@
...
           <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
         </PackageReference>
-        <PackageReference Include="Npgsql.EntityFrameworkCore.PostgreSQL" Version="8.0.2" />
+        <PackageReference Include="Npgsql.EntityFrameworkCore.PostgreSQL" Version="8.0.4" />
         <PackageReference Include="Amazon.Sqs" Version="0.33.0" />
         <PackageReference Include="AWSSDK.SQS" Version="3.7.301.31" />
         <ProjectReference Include="..\..\Libraries\CO.CDP.Functional\CO.CDP.Functional.csproj" />
         <ProjectReference Include="..\..\Libraries\CO.CDP.Swashbuckle\CO.CDP.Swashbuckle.csproj" />
         <ProjectReference Include="..\..\Libraries\CO.CDP.MQ\CO.CDP.MQ.csproj" />
+        <ProjectReference Include="..\CO.CDP.OrganisationInformation.Persistence\CO.CDP.OrganisationInformation.Persistence.csproj" />
```
jakzal, line 31: Entity verification should be an independent component. If we can, we should not introduced dependencies shared with other projects. Why do we nee OrganisationInformation.Persistence here?

Services/CO.CDP.EntityVerification/Ppon/OrganisationRegisteredEventHandler.cs
jakzal: This is something for the other story.

Services/CO.CDP.EntityVerification/Ppon/OrganisationRegisteredEventHandler.cs
```diff
@@ -1,21 +1,35 @@
...
 
-        Persistence.Ppon newIdentifier = new() { PponId = pponId };
+        List<Identifier> ids = [];
+        foreach (var e in @event.AllIdentifiers())
+        {
+            ids.Add(new Identifier { Ppon = newPpon,
+                LegalName = e.LegalName,
+                Scheme = e.Scheme,
+                Uri = e.Uri });
+        }
```
jakzal, line 28: If we hide this  mappinglogic behind a private method, the call could just be made in the field initialiser, i.e.

```csharp
Ppon newPpon = new () {
    // ...
    Identifiers = asPerisistenceIdentifiers(@event.AllIdentifiers())
}
```

This way, side effects (adding to a mutable list) would also be isolated.

Services/CO.CDP.EntityVerification/Persistence/EntityVerificationContext.cs
```diff
@@ -7,9 +8,11 @@
 
 namespace CO.CDP.EntityVerification.Persistence;
 
+
 public class EntityVerificationContext : DbContext
 {
     public DbSet<Ppon> Ppons { get; set; } = null!;
+    public DbSet<Identifier> Identifiers { get; set; } = null!;
```
jakzal, line 15: If we never query Identifiers directly, do we need a DbSet for them here? Forgive me my limitted EF Core skills.

Services/CO.CDP.EntityVerification.Tests/Persistence/DatabasePponRepositoryTest.cs
```diff
@@ -27,25 +28,15 @@ public async Task ItFindsSavedPpon()
     public void ItRejectsAnAlreadyKnownPponId()
     {
         using var repository = PponRepository();
-        var ppon = GivenPpon(pponId: "d73c9fef3c8549e2b3999fa9ec0a5aea");
-        var anotherPpon = GivenPpon(pponId: "d73c9fef3c8549e2b3999fa9ec0a5aea");
+        var ppon = PponTestHelper.GivenPpon(pponId: "d73c9fef3c8549e2b3999fa9ec0a5aea");
+        var anotherPpon = PponTestHelper.GivenPpon(pponId: "d73c9fef3c8549e2b3999fa9ec0a5aea");
```
jakzal, line 32: Great idea to extract it. Can we give it a more meaningful name (i.e. Factories), and use a static import so it keeps the test case readable (`PponTestHelper.GivenPpon` vs `GivenPpon`)?

Services/CO.CDP.EntityVerification.Tests/Persistence/IdentifierTests.cs
```diff
@@ -0,0 +1,47 @@
...
+        // Act
+        var result = EntityVerification.Persistence.Identifier.GetPersistenceIdentifiers(evIds, newPpon);
+        
+        // Assert
+        Assert.NotNull(result);
+        Assert.Single(result);
+        var identifier = result.First();
+        Assert.Equal(evIds[0].LegalName, identifier.LegalName);
+        Assert.Equal(evIds[0].Scheme, identifier.Scheme);
+        Assert.Equal(evIds[0].Uri, identifier.Uri);
```
jakzal, line 45: Let's use fluent assertions for consistency.

Services/CO.CDP.EntityVerification/CO.CDP.EntityVerification.csproj
jakzal: I think I understand this was done because of IEntityDate. To prevent the above dependency, I'm moving it out to a dedicated library: https://github.com/cabinetoffice/GCGS-Central-Digital-Platform/pull/324

Dockerfile
```diff
@@ -211,3 +211,7 @@ FROM base AS final-organisation-app
 WORKDIR /app
 COPY --from=publish-organisation-app /app/publish .
 ENTRYPOINT ["dotnet", "CO.CDP.OrganisationApp.dll"]
+
+FROM localstack/localstack:3.5 AS localstack
+COPY --chown=localstack ./localstack-init/create-queues.sh /etc/localstack/init/ready.d/init-aws.sh
+RUN chmod +x /etc/localstack/init/ready.d/init-aws.sh
```
jakzal, line 217: So far, we've only been using this Dockerfile for production dotnet services. I'd like to keep it like that. Localstack is not only not a production service, but entirely unrelated to our dotnet services (doesn't share any build steps). 

I think in this case it's not even necessary to create a dedicated image. Just mount the ready.d volume from `./docker/localstack/ready.d`.

compose.yml
```diff
@@ -50,10 +50,17 @@ services:
...
     healthcheck:
-      test: [ "CMD-SHELL", "bash", "-c", "awslocal sqs list-queues && awslocal s3 ls"]
+      test: ["CMD-SHELL", "awslocal sqs list-queues --region eu-west-1 | grep -E 'ev-inbound|ev-outbound' || exit 1"]
+      interval: 5s
+      timeout: 5s
+      retries: 5
     environment:
       # LocalStack configuration: https://docs.localstack.cloud/references/configuration/
-      DEBUG: 0
+      - SERVICES=sqs
```
jakzal, line 59: We also need s3

compose.yml
```diff
@@ -235,12 +242,20 @@ services:
...
       ASPNETCORE_ENVIRONMENT: Production
       ConnectionStrings__EvDatabase: 'Server=db;Database=cdp;Username=cdp_user;Password=cdp123;'
-      Sqs__ServiceURL: "http://localstack:4566"
\ No newline at end of file
+      Sqs__ServiceURL: "http://localstack:4566"
+      Sqs__AuthenticationRegion: "eu-west-1"
+      InboundQueue__Name: "ev-inbound"
+      InboundQueue__MaxNumberOfMessages: 10
+      InboundQueue__WaitTimeSeconds: 20
+      OutboundQueue__Name: "ev-outbound"
```
jakzal, line 261: Only `Sqs__ServiceURL` is necessary here, as this the only different setting between appsettings.Development.json and docker.

docker/localstack/ready.d/init-aws.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+echo "Waiting for localstack to be ready...."
+sleep 5
```
jakzal, line 4: Is this really needed? I'd think localstack runs its ready.d scripts when it's up?

jakzal: Looks like it isn't needed. By the time this script is executed, localstack is ready: https://docs.localstack.cloud/references/init-hooks/

```
/etc
└── localstack
    └── init
        ├── boot.d           <-- executed in the container before localstack starts
        ├── ready.d          <-- executed when localstack becomes ready
        ├── shutdown.d       <-- executed when localstack shuts down
        └── start.d          <-- executed when localstack starts up
```

PR #352, opened 2024-07-25
DP-214 - update roles within the stub api to be lower case

maciej-goaco:


maciej-goaco:
Replaced with https://github.com/cabinetoffice/GCGS-Central-Digital-Platform/pull/353

PR #368, opened 2024-07-26
Feature/dp 389 ppon service adds identifiers for organisation updated event

dbgoaco:
PPON Service - PPON Service adds list of identifiers in response to Organisation Updated Event


glgoaco (approved):

Services/CO.CDP.EntityVerification.Tests/Events/EventDeserializerTest.cs
```diff
@@ -58,4 +69,28 @@ public void ItThrowsAnExceptionIfTheEventCannotBeDeserialized()
             ],
             Roles = ["supplier"]
         };
+
+    private static OrganisationUpdated GvienOrganisationUpdatedEvent() =>
+        new()
+        {
+            PponId = "2e1230b94ced4fa9aac76895f122e9f0",
```
jakzal, line 76: We discussed the `OrganisationUpdated` event should have same fields as `OrganisationRegistered` (both based on OCDS). We won't have a `PponId` in there. PPON is just one of the identifiers.

https://github.com/cabinetoffice/GCGS-Central-Digital-Platform/blob/5a4a7d77410d5eff281d28426605795c120ecbc9/Services/CO.CDP.Organisation.WebApi/Events/OrganisationUpdated.cs#L3-L20

jakzal (changes requested):

Services/CO.CDP.EntityVerification.Tests/Events/EventDeserializerTest.cs
jakzal: In the EntityVerification project you can just copy the `OrganisationRegistered` and rename it to `OrganisationUpdated`.

https://github.com/cabinetoffice/GCGS-Central-Digital-Platform/blob/5a4a7d77410d5eff281d28426605795c120ecbc9/Services/CO.CDP.EntityVerification/Events/InboundEvents.cs#L5-L20

dbgoaco: Done.

Services/CO.CDP.EntityVerification/Events/Values.cs
```diff
@@ -2,6 +2,8 @@ namespace CO.CDP.EntityVerification.Events;
 
 public record Identifier
 {
+    public const string PponSchemeName = "CDP-PPON";
```
jakzal, line 5: This value is used with Events. If an event is serialised, would this constant be also serialised? 

If the answer is yes, then it shouldn't be here :)

Services/CO.CDP.EntityVerification/Ppon/OrganisationUpdatedSubscriber.cs
```diff
@@ -0,0 +1,38 @@
...
+            var pponToUpdate = await pponRepository.FindPponByPponIdAsync(pponIdentifier.Id);
+
+            if (pponToUpdate != null)
+            {
+                var identifiersToPersist = Persistence.Identifier.GetPersistenceIdentifiers(@event.AllIdentifiers());
+
+                foreach (var identifier in identifiersToPersist)
+                {
+                    pponToUpdate.Identifiers.Add(identifier);
+                }
```
jakzal, line 32: Since the `OrganisationUpdated` event always contains all the latest organisation details, the subscriber is responsible for reconciling changes. I know it's not the usual case for events, but that's what we went for by deciding to publish coarse events with most organisation details. If we don't check which identifiers were already here, we will be appending them every time.

dharmverma (approved):

Services/CO.CDP.EntityVerification/Events/Values.cs
dbgoaco: Good point. Should we define the possible schemes in a JSON file and load it in?

jakzal: There's a task for this in the backlog :)

jakzal: For now, maybe let's define a dedicated class with constants? i.e `IdentifierSchemes.Ppon`

dbgoaco: Done.

Services/CO.CDP.EntityVerification/Ppon/OrganisationUpdatedSubscriber.cs
dbgoaco: Now only adds Identifiers that do not exist.

Services/CO.CDP.EntityVerification.Tests/Ppon/PponServiceTests.cs
```diff
@@ -1,6 +1,6 @@
 using CO.CDP.EntityVerification.Ppon;
 
-namespace CO.CDP.EntityVerification.Tests.Ppon;
+namespace CO.CDP.EntityVerification.Tests.Events;
```
jakzal, line 3: Seems unintended. As many files in the Tests/Ppon directory.

jakzal: Fixed.

jakzal (approved):

PR #370, opened 2024-07-27
DP-191 Add Pipeline to update accounts

webit4me:


terragrunt/modules/ecs/templates/task-definitions/organisation.json.tftpl
```diff
@@ -22,6 +22,8 @@
     "environment": [
         {"name": "ASPNETCORE_ENVIRONMENT", "value": "${environment}"},
         {"name": "ASPNETCORE_HTTP_PORTS", "value": "${host_port}"},
+        {"name": "Aws__SqsDispatcher__QueueUrl", "value": "${queue_inbound_url}"},
+        {"name": "Aws__SqsPublisher__QueueUrl", "value": "${queue_inbound_url}"},
```
jakzal, line 26: Heads up: queues that are inbound for one service are outbound for another.i.e. Inbound queue for entity verification is an outbound queue for the organisation service. It might be confusing but we didn’t come up with better names.

jakzal (approved):

terragrunt/modules/ecs/templates/task-definitions/organisation.json.tftpl
webit4me: Thanks, it is confusing indeed. I will amend naming in the platform to make it more understandable at least this side

PR #396, opened 2024-07-30
DP-200: Add feature flag for SwaggerUI in Services

mareg:
A simple feature flag implemented with `appsettings.*.json`:
```json
"Features": {
    "SwaggerUI": false
}
```

SwaggerUI has been enabled on `Development` builds and disabled on production builds.

### DIFF
--- a/Services/CO.CDP.Tenant.WebApi/Program.cs
+++ b/Services/CO.CDP.Tenant.WebApi/Program.cs
@@ -48,12 +48,13 @@ var app = builder.Build();
 app.UseForwardedHeaders();
 
 // Configure the HTTP request pipeline.
-if (app.Environment.IsDevelopment())
+if (builder.Configuration.GetValue("Features:SwagerUI", false))
 {
     app.UseSwagger();
     app.UseSwaggerUI();
 }
-else
+
+if (!app.Environment.IsDevelopment())
 {
     app.UseExceptionHandler();
     app.UseHsts();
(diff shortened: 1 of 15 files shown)


Services/CO.CDP.Tenant.WebApi/Program.cs
```diff
@@ -15,10 +15,16 @@
 var builder = WebApplication.CreateBuilder(args);
 builder.ConfigureForwardedHeaders();
 
+var enableSwagger = builder.Configuration.GetSection("Features").GetValue<bool>("SwaggerUI");
```
jakzal, line 18: Use the `builder.Configuration.GetValue("Features:SwagerUI", false)` syntax please, as it won’t throw an exception if the section isn’t defined but will fallback to the default value instead.

Services/CO.CDP.Tenant.WebApi/Program.cs
```diff
@@ -15,10 +15,16 @@
...
+
 // Add services to the container.
 // Learn more about configuring Swagger/OpenAPI at https://aka.ms/aspnetcore/swashbuckle
 builder.Services.AddEndpointsApiExplorer();
-builder.Services.AddSwaggerGen(options => { options.DocumentTenantApi(); });
+if (enableSwagger)
+{
+    builder.Services.AddSwaggerGen(options => { options.DocumentTenantApi(); });
+}
```
jakzal, line 26: This call should not be behind a feature flag afaik.

jakzal: This is what broke the tests by the way.

jakzal (approved):

webit4me (approved):

PR #405, opened 2024-07-31
DP-200: Enable SwaggerUI feature flag override for Aws envs

mareg:


jakzal (approved):

jakzal (changes requested):
Authority service exposes swagger too. It needs to be behind a feature flag and it needs to have aws environment configuration files as well.

webit4me (approved):

jakzal (approved):

PR #409, opened 2024-07-31
DP-396: Do not crash the dispatcher loop in case of an exception

jakzal:
This way, in case of a problem, the background service will continue running.

### DIFF
--- a/Libraries/CO.CDP.AwsServices/Sqs/SqsDispatcher.cs
+++ b/Libraries/CO.CDP.AwsServices/Sqs/SqsDispatcher.cs
@@ -58,13 +58,25 @@ public class SqsDispatcher(
 
             while (!cancellationToken.IsCancellationRequested)
             {
-                foreach (var message in await ReceiveMessagesAsync(configuration.QueueUrl, cancellationToken))
+                try
                 {
-                    await HandleMessage(message);
-                    await DeleteMessage(message);
+                    await HandleMessages(cancellationToken);
+                }
+                catch (Exception cause)
+                {
+                    logger.LogError(cause, "Message handling failed");
                 }
             }
         }, cancellationToken);
+    }
+
+    private async Task HandleMessages(CancellationToken cancellationToken)
+    {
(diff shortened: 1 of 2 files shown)


Libraries/CO.CDP.AwsServices/Sqs/SqsDispatcher.cs
```diff
@@ -58,15 +58,27 @@ public Task ExecuteAsync(CancellationToken cancellationToken = default)
...
             }
         }, cancellationToken);
     }
 
+    private async Task HandleMessages(CancellationToken cancellationToken)
+    {
+        foreach (var message in await ReceiveMessagesAsync(configuration.QueueUrl, cancellationToken))
+        {
+            await HandleMessage(message);
+            await DeleteMessage(message);
```
jakzal, line 78: Consider wrapping these two lines in a try/catch block. Otherwise, if there's more than one message received, and one of their handlers fails, the whole batch will be retried.

jakzal: To be precise, not a whole batch will be retried, but what was remaining. The problem is that some newer messages might be handled before those to be retried come back to the pool.

oh, but  messages  past the failed one will be discarded for now




Example:

* ReceiveMessages gets 10 messages (id 1-10)
  * HandleMessage(1)
  * DeleteMessage(1)
  * HandleMessage(2)
  * DeleteMessage(2)
  * HandleMessage(3) <- throws an exception and processing goes out of the foreach loop
* Exception is caught and continues to fetch new messages
* ReceiveMessages gets 10 new messages (id 11-20)
  * HandleMessage(11)
  * DeleteMessage(11)
  * ...
  * HandleMessage(20)
  * DeleteMessage(20)
* ReceiveMessages gets 10 messages (id 3-10, grace period passed, these messages were not removed so came back to the pool)

jakzal: Updated.

dpatel017 (approved):

glgoaco (approved):

PR #415, opened 2024-08-01
DP-408

webit4me:
- Add Module to provision Queues and their DLQs
- Grant Services to permission to connect to all queues using task IAM role
- Set maxReceiveCount to 10
- Update health-check service to use the new DLQs


terragrunt/modules/sqs/variables.tf
```diff
@@ -0,0 +1,55 @@
...
+variable "max_receive_count" {
+  default     = 10
+  description = ""
+  type        = number
+}
+
+variable "message_retention_seconds" {
+  description = "The number of seconds Amazon SQS retains a message."
+  type        = number
+  default     = 345600 # 4 Days
```
jakzal, line 22: We shouldn't delete until the message is consumed.

webit4me: It's impossible, The max amount of time you can hold on to it is 2 weeks. 
Hence existence of DLQs 
[4 days is AWS default, which seems reasonable in most designs](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_SetQueueAttributes.html).

dharmverma (approved):

PR #445, opened 2024-08-09
Initial working version of Supplier sees their PPON if their organisa…

dbgoaco:
[DP-387](https://noticingsystem.atlassian.net/jira/software/projects/DP/boards/22?selectedIssue=DP-387)

### DIFF
--- a/Frontend/CO.CDP.OrganisationApp/Pages/OrganisationOverview.cshtml
+++ b/Frontend/CO.CDP.OrganisationApp/Pages/OrganisationOverview.cshtml
@@ -6,8 +6,9 @@
 @{
     var organisationDetails = Model.OrganisationDetails!;
     var registeredAddress = organisationDetails.Addresses.FirstOrDefault(a => a.Type == Organisation.WebApiClient.AddressType.Registered);
-    var pponIndentifier = @organisationDetails.AdditionalIdentifiers.First(i => i.Scheme == "CDP-PPON");
-    var showPpon = pponIndentifier != null && organisationDetails.Identifier.Scheme == "Other";
+    var pponIndentifier = organisationDetails.AdditionalIdentifiers.Prepend(organisationDetails.Identifier).FirstOrDefault(i => i.Scheme == "CDP-PPON");
+    var otherIdentifier = organisationDetails.AdditionalIdentifiers.Prepend(organisationDetails.Identifier).FirstOrDefault(i => i.Scheme == "Other");
+    var showPpon = pponIndentifier != null && otherIdentifier != null;
 }
 
 <main class="govuk-main-wrapper">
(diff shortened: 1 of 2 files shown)


dpatel017 (approved):

Frontend/CO.CDP.OrganisationApp/Pages/OrganisationOverview.cshtml
```diff
@@ -6,6 +6,9 @@
 @{
     var organisationDetails = Model.OrganisationDetails!;
     var registeredAddress = organisationDetails.Addresses.FirstOrDefault(a => a.Type == Organisation.WebApiClient.AddressType.Registered);
+    var pponIndentifier = organisationDetails.AdditionalIdentifiers.Prepend(organisationDetails.Identifier).FirstOrDefault(i => i.Scheme == "CDP-PPON");
+    var otherIdentifier = organisationDetails.AdditionalIdentifiers.Prepend(organisationDetails.Identifier).FirstOrDefault(i => i.Scheme == "Other");
```
jakzal, line 10: This isn't quite right.

We should only show PPON if it's the only identifier we have (ignoring "Other").

Otherwise, we should show the primary identifier here.

jakzal: Oh, I guess that's what happens! I misunderstood "otherIdentifier"'s purpose.

Frontend/CO.CDP.OrganisationApp/Pages/OrganisationOverview.cshtml
```diff
@@ -35,9 +38,25 @@
...
                             <p class="govuk-body">
-                                @organisationDetails.Identifier.Scheme.SchemeDescription()
+                                @if (showPpon)
+                                {
+                                    @pponIndentifier!.Scheme.SchemeDescription()
+                                }
+                                else
+                                {
+                                    @organisationDetails.Identifier.Scheme.SchemeDescription()
+                                }
```
jakzal, line 48: Instead, of using `showPpon` can we define an `identifier` variable to show it here? This way we wouldn't need conditions in the template.

Frontend/CO.CDP.OrganisationApp/Pages/OrganisationOverview.cshtml
dbgoaco: It's ok, I thought Sammy wanted us to show a PPON if there was an Other?

Frontend/CO.CDP.OrganisationApp/Pages/OrganisationOverview.cshtml
dbgoaco: Done.

jakzal (approved):

PR #446, opened 2024-08-09
DP-455: Add an unique index constraint to the API Key

mareg:
Adds a unique constraint to the `key`.


mareg:
@dharmverma do we want the `name` to have a unique constraint as well?

glgoaco (approved):

dharmverma (approved):

PR #472, opened 2024-08-18
Authority - Refresh tokens

dharmverma:


Services/CO.CDP.Organisation.Authority/Endpoint.cs
```diff
@@ -66,6 +66,14 @@ public static void UseIdentity(this WebApplication app)
                         }
                         return Results.Problem(statusCode: StatusCodes.Status400BadRequest, detail: "Invalid client secret");
 
+                    case GrantTypes.RefreshToken:
+                        (valid, urn) = await service.ValidateRefreshToken(request.RefreshToken);
+                        if (valid)
+                        {
+                            return Results.Ok(await service.CreateToken(urn!));
```
jakzal, line 73: Wouldn't it be better to check if `urn != null` in the above condition, rather than use `urn!`? This way, it would fallback to returning a problem response, rather than a runtime exception.

Frontend/CO.CDP.OrganisationApp/ApiBearerTokenHandler.cs
```diff
@@ -11,58 +11,101 @@ public class ApiBearerTokenHandler(
...
     }
 
-    private async Task<string> GetAccessTokenAsync(string oneLogintoken)
+    private async Task<string> GetOneloginAccessToken()
+    {
+        var context = httpContextAccessor.HttpContext;
+        if (context != null)
+        {
+            var oneLogintoken = await context.GetTokenAsync("access_token");
+            // var expiresAt = DateTimeOffset.Parse(await context.GetTokenAsync("expires_at")).ToLocalTime();
```
jakzal, line 48: Any reason to keep this comment here?

jakzal (approved):

dpatel017 (approved):

PR #524, opened 2024-08-28
Changed FormType and Scope to Enum

shilpigoel-goaco:


Services/CO.CDP.DataSharing.WebApi/Model/Form.cs
```diff
@@ -36,8 +51,8 @@ public record Form
     public required string ShareCode { get; init; }
 
     /// <example>0</example>
-    public required int Scope { get; init; }
+    public required FormScope Scope { get; init; }
 
     /// <example>0</example>
-    public required int Type { get; init; }
+    public required FormSectionType Type { get; init; }
```
jakzal, line 57: This should be defined on the section rather than form, since we moved it there.

Services/CO.CDP.DataSharing.WebApi/Model/Form.cs
```diff
@@ -36,8 +51,8 @@ public record Form
     public required string ShareCode { get; init; }
 
     /// <example>0</example>
```
jakzal, line 53: Examples should be updated. I think if you remove it alltogether, it will be automatically generated from the enum, but you'll need to confirm.

Services/CO.CDP.DataSharing.WebApi/Model/Form.cs
shilpigoel-goaco: there is no FormSection Model in the DataSharing API.

jakzal (approved):

PR #571, opened 2024-09-05
DP-527 - Prevent empty VAT from being persisted

dbgoaco:
[DP-527](https://noticingsystem.atlassian.net/jira/software/projects/DP/boards/22?selectedIssue=DP-527)


dharmverma (approved):

Services/CO.CDP.Organisation.WebApi.Tests/UseCase/UpdateOrganisationUseCaseTest.cs
```diff
@@ -202,6 +202,40 @@ public async Task Execute_ShouldPublishOrganisationUpdatedEvent_WhenOrganisation
...
+        var command = new UpdateOrganisation
+        {
+            Type = OrganisationUpdateType.AdditionalIdentifiers,
+            Organisation = new OrganisationInfo
+            {
+                AdditionalIdentifiers = [new OrganisationIdentifier
+                {
+                    Id = "",
+                    LegalName = "Acme",
+                    Scheme = "GB-VAT-EMPTY"
```
jakzal, line 217: This isn’t  ascheme we expect. It’s just “VAT” I think.

dbgoaco: Done.

Services/CO.CDP.Organisation.WebApi.Tests/UseCase/UpdateOrganisationUseCaseTest.cs
```diff
@@ -202,6 +202,39 @@ public async Task Execute_ShouldPublishOrganisationUpdatedEvent_WhenOrganisation
...
+        _organisationRepositoryMock.Setup(repo => repo.Find(_organisationId)).ReturnsAsync(organisation);
+
+        var result = await UseCase.Execute((_organisationId, command));
+
+        result.Should().BeTrue();
+        _organisationRepositoryMock.Verify(repo => repo.Save(organisation!), Times.Once);
+
+        organisation.Identifiers.FirstOrDefault(i => i.Scheme == "GB-VAT").Should().BeNull();
+    }
+
```
jakzal, line 237: This only handles the case of adding an identifier. What about the case when someone adds VAT but then changes their answer?

Services/CO.CDP.Organisation.WebApi.Tests/UseCase/UpdateOrganisationUseCaseTest.cs
dbgoaco: Done.

Services/CO.CDP.Organisation.WebApi.Tests/UseCase/UpdateOrganisationUseCaseTest.cs
jakzal: I mean the situation when someone gives the VAT number at first:
image

But then, they change their answer to "No, I don't have a VAT number":

image

jakzal (changes requested):

Services/CO.CDP.Organisation.WebApi.Tests/UseCase/UpdateOrganisationUseCaseTest.cs
dbgoaco: Added Execute_ShouldUpdateVatNumber_WhenVatIdentifierAlreadyExists

jakzal:
Now, if I select "No", the answer is not persisted as the supplier information screen doesn't know I answered the question:

image

It seems that `completed_vat` flag is not set in `supplier_information` table.

Services/CO.CDP.Organisation.WebApi/UseCase/UpdateOrganisationUseCase.cs
```diff
@@ -116,4 +123,35 @@ public async Task<bool> Execute((Guid organisationId, UpdateOrganisation updateO
...
+        {
+            nextPrimaryIdentifier = organisation.Identifiers.FirstOrDefault(i =>
+                i.Scheme == AssignIdentifierUseCase.IdentifierSchemes.Ppon);
+
+            if (nextPrimaryIdentifier == null)
+            {
+                nextPrimaryIdentifier = organisation.Identifiers.FirstOrDefault(i =>
+                    i.Scheme == AssignIdentifierUseCase.IdentifierSchemes.Other);
+            }
+        }
```
jakzal, line 146: I'm not sure about this code. It only seems to be looking at PPON and Other type of identifiers. Doesn't seem to be right.

dbgoaco: Now checks whether the primary identifier is being removed. If it is, it finds the next primary identifier, if there is one.

jakzal (approved):

PR #608, opened 2024-09-12
DP-222 Name env_var after technology not componenet

webit4me:
- Currect services env_var names


mareg (approved):

webit4me:
closing this to be addressed if/when we have capacity

PR #615, opened 2024-09-13
Api key management

dpatel017:


Services/CO.CDP.Organisation.WebApi/Api/Organisation.cs
```diff
@@ -645,6 +645,85 @@ await useCase.Execute((organisationId, personInviteId))
 
         return app;
     }
+
+    public static RouteGroupBuilder UseManageApiKeyEndpoints(this RouteGroupBuilder app)
+    {
+        app.MapGet("/{organisationId}/api-keys",
+            async (Guid organisationId, IUseCase<Guid, IEnumerable<Model.AuthenticationKey>> useCase) =>
+               await useCase.Execute(organisationId)
+                   .AndThen(entities => entities != null ? Results.Ok(entities) : Results.NotFound()))
```
jakzal, line 654: Can `entities` actually be `null`, or just an empty collection?

Services/CO.CDP.Organisation.WebApi/Api/Organisation.cs
```diff
@@ -645,6 +645,85 @@ await useCase.Execute((organisationId, personInviteId))
...
+               operation.Responses["500"].Description = "Internal server error.";
+               return operation;
+           });
+
+        app.MapPost("/{organisationId}/api-keys",
+            async (Guid organisationId, RegisterAuthenticationKey registerAuthenticationKey,
+                IUseCase<(Guid, RegisterAuthenticationKey), bool> useCase) =>
+
+                await useCase.Execute((organisationId, registerAuthenticationKey))
+                    .AndThen(_ => Results.Ok())
```
jakzal, line 678: When a new resource is created, and we do not want to return the response body, we should use `201 Created` with a `Location` header pointing to the created resource.

In your case, I do not think we have an endpoint to return a single API key, so we cannot return the `Location` header, but I think we should still return 201 created rather than 200.

Services/CO.CDP.Organisation.WebApi/Model/Command.cs
```diff
@@ -336,12 +336,27 @@ public bool TryGetName(out string name)
...
+    public Guid OrganisationId { get; set; }
+    public bool? Revoked { get; set; }
+}
+
+public record AuthenticationKey
+{
+    public required string Name { get; set; }
+    public bool? Revoked { get; set; }
+    public DateTimeOffset CreatedOn { get; set; }
+    public DateTimeOffset UpdatedOn { get; set; }
```
jakzal, line 352: As a rule of thumb CreatedOn/UpdatedOn properties should not be exposed to clients. Do we need them in the frontend? What for?

Services/CO.CDP.OrganisationInformation.Persistence/AuthenticationKey.cs
```diff
@@ -13,6 +12,7 @@ public class AuthenticationKey : IEntityDate
 
     [ForeignKey(nameof(Organisation))]
     public int? OrganisationId { get; set; }
+    public bool? Revoked { get; set; }
```
jakzal, line 15: I'd suggest making it non-null, default `false`. This way we only have two states to deal with. What would `null` mean in this context anyway?

jakzal (approved):
Approved, but left few comments to consider before the merge.

Services/CO.CDP.Organisation.WebApi/Model/Command.cs
dpatel017: This fields used to populate date created and date cancelled. 
image

Services/CO.CDP.Organisation.WebApi/Api/Organisation.cs
dpatel017: It will be just empty collection

jakzal: So why is there a null check here?

Services/CO.CDP.OrganisationInformation.Persistence/AuthenticationKey.cs
dpatel017: Well created null because it was existing table.  Will make it default false.

JBaigGoaco (approved):

Services/CO.CDP.Organisation.WebApi/Api/Organisation.cs
dpatel017: removed null check

Services/CO.CDP.Organisation.WebApi/Api/Organisation.cs
dpatel017: Done

Services/CO.CDP.Organisation.WebApi/Model/Command.cs
dpatel017: Done

PR #618, opened 2024-09-16
Feature/dp 541 declaring no exclusions

JBaigGoaco:
declaring no exclusions page with validations

image


Frontend/CO.CDP.OrganisationApp/Pages/Exclusions/DeclaringExclusions.cshtml
```diff
@@ -0,0 +1,54 @@
...
+
+                        <div class="govuk-radios" data-module="govuk-radios">
+                            <div class="govuk-radios__item">
+                                <input class="govuk-radios__input" id="conditional-YES" name="YesNoInput" type="radio"
+                                       value="true" data-aria-controls="conditional-YES" @(Model.YesNoInput == true ? "checked" : "")>
+                                <label class="govuk-label govuk-radios__label" for="conditional-YES">Yes</label>
+                            </div>
+                            <div class="govuk-radios__item">
+                                <input class="govuk-radios__input" id="conditional-NO" name="YesNoInput" type="radio"
+                                       value="false" data-aria-controls="conditional-NO" @(Model.YesNoInput == false ? "checked" : "")>
```
andymantell, line 39: `data-aria-controls` attributes are used by the govuk JavaScript to control conditionally revealing content (See https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/govuk/components/radios/radios.mjs#L21)

That doesn't seem to be the case here so these could be removed from both inputs

andymantell (changes requested):

Frontend/CO.CDP.OrganisationApp/Pages/Exclusions/DeclaringExclusions.cshtml
JBaigGoaco: Thank you Andy - thats a very good point. I have removed it.

dharmverma (approved):

dpatel017 (approved):

andymantell (approved):

PR #627, opened 2024-09-18
DP-193 - Mapping Schemes to URIs added

maciej-goaco:


Services/CO.CDP.OrganisationInformation/Constants.cs
```diff
@@ -0,0 +1,30 @@
+namespace CO.CDP.OrganisationInformation;
+public static class Constants
```
jakzal, line 2: Let's name it with something that reflects what it's for, i.e. `IdentifierSchemes`.

Services/CO.CDP.OrganisationInformation/Constants.cs
```diff
@@ -0,0 +1,30 @@
...
+        {"GB-NHS", "https://odsportal.digital.nhs.uk/Organisation/OrganisationDetails?organisationId={0}"},
+        {"GB-UKPRN", "https://www.ukrlp.co.uk"},
+        {"GB-MPR", "https://mutuals.fca.org.uk/Search/Society/{0}"},
+        {"GG-RCE", "https://portal.guernseyregistry.com/e-commerce/company/{0}"},
+        {"JE-FSC", "https://www.jerseyfsc.org/registry/registry-entities/entity/{0}"},
+        {"IM-CR", "https://services.gov.im/ded/services/companiesregistry/companysearch.iom"},
+        {"GB-PPON", "https://find-and-update.company-information.service.gov.uk/company/{0}"}
+    };
+
+    public static Uri? GetEndpointUri(string? scheme, string? identifierId)
```
jakzal, line 19: How about `GetRegistryUri`?

maciej-goaco: done

Services/CO.CDP.OrganisationInformation/Constants.cs
maciej-goaco: done

dharmverma (commented):
There are few GitHub warning, can that be resolved?

Services/CO.CDP.DataSharing.WebApi.Tests/AutoMapper/DataSharingProfileTest.cs
```diff
@@ -5,6 +5,6 @@ public class DataSharingProfileTest(AutoMapperFixture mapperFixture) : IClassFix
     [Fact]
     public void ConfigurationIsValid()
     {
-        mapperFixture.Configuration.AssertConfigurationIsValid();
+        mapperFixture.Configuration?.AssertConfigurationIsValid();
```
jakzal, line 8: Seems wrong. If the Configuration is null, the test should fail. Otherwise, there's no point of it.

Services/CO.CDP.DataSharing.WebApi.Tests/UseCase/GetSharedDataUseCaseTest.cs
```diff
@@ -115,10 +117,11 @@ private void AssertAdditionalIdentifiers(IEnumerable<Identifier>? additionalIden
         additionalIdentifiers.Should().NotBeNull();
         additionalIdentifiers.Should().HaveCount(1);
 
+        var scheme = "GB-COH";
         var identifier = additionalIdentifiers?.First();
-        identifier?.Scheme.Should().Be("GB-COH");
+        identifier?.Scheme.Should().Be(scheme);
         identifier?.LegalName.Should().Be("AnotherLegalName");
-        identifier?.Uri.Should().Be(new Uri("http://example.com"));
+        identifier?.Uri.Should().Be(IdentifierSchemes.GetRegistryUri(_configurationService.Object.GetOrganisationApiHostUrl(), scheme, identifier.Id));
```
jakzal, line 124: `configurationService.Object.GetOrganisationApiHostUrl()` is never set up, so it always returns an empty string? It's also not passed anywhere, so isn't actually used by code under test, or am i missing something?

Services/CO.CDP.DataSharing.WebApi/Program.cs
```diff
@@ -10,23 +11,35 @@
...
 // Learn more about configuring Swagger/OpenAPI at https://aka.ms/aspnetcore/swashbuckle
 builder.Services.AddEndpointsApiExplorer();
 builder.Services.AddSwaggerGen(options => { options.DocumentDataSharingApi(builder.Configuration); });
 builder.Services.AddHealthChecks()
     .AddNpgSql(ConnectionStringHelper.GetConnectionString(builder.Configuration, "OrganisationInformationDatabase"));
-builder.Services.AddAutoMapper(typeof(DataSharingProfile));
+
+builder.Services.AddScoped(provider => new MapperConfiguration(cfg =>
+{
+    var configurationService = provider.GetService<IConfigurationService>();
```
jakzal, line 33: I'd rather use `GetRequiredService` and rely on it never being `null`.

Services/CO.CDP.DataSharing.WebApi.Tests/AutoMapper/DataSharingProfileTest.cs
maciej-goaco: That's true, makes no point as won't happen as it would actually fail much earlier if unable to load the consumed setting.
However, made it this way only to satisfy GitHub complaints, so opened for any suggestions.

PR #630, opened 2024-09-19
DP-612 - CDP-PPON updated to GB-PPON

maciej-goaco:


dbgoaco (commented):
I think we need a migration to update existing identifier records in EV and Sirsi databases so that existing identifiers are also updated from CDP-PPON to GB-PPON

dbgoaco (commented):
Looks good but also need the same for entity_verification.identifiers as it also stores the scheme

dbgoaco (approved):

Services/CO.CDP.OrganisationInformation.Persistence/Migrations/20240919141551_UpdatePponSchemeName.cs
```diff
@@ -0,0 +1,22 @@
...
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.Sql("UPDATE identifiers SET scheme = 'GB-PPON' WHERE scheme = 'CDP-PPON';");
+        }
+
+        /// <inheritdoc />
+        protected override void Down(MigrationBuilder migrationBuilder)
+        {
+
```
dharmverma, line 19: Add migration down script, for just in case scenarios :)

dharmverma (approved):

dharmverma (approved):

jakzal (approved):

Services/CO.CDP.OrganisationInformation.Persistence/Migrations/20240919141551_UpdatePponSchemeName.cs
maciej-goaco: done

PR #633, opened 2024-09-19
DP-613: Add "Name" field to the FormQuestion entity

mareg:


JBaigGoaco (approved):

mareg:
screenshot2371

Tested with a fresh DB that migrations populate the column.

dbgoaco (approved):

PR #635, opened 2024-09-20
Migration for dp 260 support admin - organisation approval

rmohammed-goaco:


shilpigoel-goaco (approved):

JBaigGoaco (approved):

Services/CO.CDP.OrganisationInformation.Persistence/Organisation.cs
```diff
@@ -21,6 +21,9 @@ public class Organisation : IEntityDate
     public BuyerInformation? BuyerInfo { get; set; }
     public DateTimeOffset CreatedOn { get; set; }
     public DateTimeOffset UpdatedOn { get; set; }
+    public DateTimeOffset? ApprovedOn { get; set; }
+    public Person? ApprovedBy { get; set; }
```
dharmverma, line 25: As we discussed earlier in dev tech discussion, can we have original column property as well along with above navigational property
`public int? ApprovedById { get; set; }`

rmohammed-goaco: Thanks. Have added this now.

PR #637, opened 2024-09-20
Feature/dp 260 support admin landing page

rmohammed-goaco:
Landing page for support admin. 
- Allows switching between buyer and supplier organisations
- Pagination will be added in a later update
- The approval functionality will also be in a separated PR
- Locking down to the SUPPORTADMIN role will be in a separate PR

24/09/2024
Agreed with @jakzal to refactor an existing endpoint `GET organisations` to get all organisations and include the relevant updated fields. The endpoint no longer requires passing in the `userUrn` and new `Details` field has been added to contain the `Approval` data and any future additions. 

The previously added admin specific endpoint has been removed.


Frontend/CO.CDP.OrganisationApp.Tests/CO.CDP.OrganisationApp.Tests.csproj
```diff
@@ -34,4 +34,15 @@
...
+    <Content Update="Pages\Support\Organisations.cshtml">
+      <ExcludeFromSingleFile>true</ExcludeFromSingleFile>
+      <CopyToPublishDirectory>PreserveNewest</CopyToPublishDirectory>
+    </Content>
+  </ItemGroup>
+
+  <ItemGroup>
+    <AdditionalFiles Include="Pages\Support\Organisations.cshtml" />
+  </ItemGroup>
+
```
rmohammed-goaco, line 47:
```suggestion
```

Frontend/CO.CDP.OrganisationApp/Pages/Support/Organisations.cshtml.cs
```diff
@@ -0,0 +1,28 @@
...
+    ISession session) : LoggedInUserAwareModel(session)
+{
+    [BindProperty]
+    public string? Title { get; set; }
+
+    [BindProperty]
+    public string? Type { get; set; }
+
+    [BindProperty]
+    public IList<ApprovableOrganisation> Organisations { get; set; } = [];
```
andymantell, line 17: You shouldn't need model binding to happen on Title and Organisations from what I can see of your code - you don't want these to be populated from request data since you're assigning values to them yourself?

You could keep the `[BindProperty(SupportsGet = true)]` on Type if you wanted to tell it to bind from the query string. But you don't need to because you've got string type as an argument on the OnGet method

```suggestion
    public string? Title { get; set; }

    public string? Type { get; set; }

    public IList<ApprovableOrganisation> Organisations { get; set; } = [];
```

andymantell (changes requested):

Frontend/CO.CDP.OrganisationApp/Pages/Support/Organisations.cshtml.cs
rmohammed-goaco: Thanks for the feedback. I've applied those changes and tested they work.

andymantell (approved):

Services/CO.CDP.Organisation.WebApi/Api/Organisation.cs
```diff
@@ -22,6 +22,24 @@ public static class EndpointExtensions
 {
     public static void UseOrganisationEndpoints(this WebApplication app)
     {
+        app.MapGet("/admin/organisations",
```
jakzal, line 25: I think we should just be using the `/organisations` endpoint(s) here.

jakzal (changes requested):

Services/CO.CDP.Organisation.WebApi/Api/Organisation.cs
```diff
@@ -32,8 +50,8 @@ await useCase.Execute(userUrn)
             .WithOpenApi(operation =>
             {
                 operation.OperationId = "ListOrganisations";
-                operation.Description = "Get a list of organisations.";
-                operation.Summary = "Get a list of organisations.";
+                operation.Description = "Get a list of organisations for a particular userUrn.";
+                operation.Summary = "Get a list of organisations for a particular userUrn.";
```
jakzal, line 54: Placing the `userUrn` here was a mistake. It was to accommodate the organisation selection screen, which was wrongly using this endpoint. It was recently fixed and the organisation selection screen no longer uses this endpoint. You should be using it instead of introducing the new `/admin/organisations`.

Services/CO.CDP.Organisation.WebApi/Api/Organisation.cs
```diff
@@ -22,6 +22,24 @@ public static class EndpointExtensions
 {
     public static void UseOrganisationEndpoints(this WebApplication app)
     {
+        app.MapGet("/admin/organisations",
+                async ([FromQuery] string type, [FromQuery] int limit, [FromQuery] int skip, IUseCase<(string, int, int), IEnumerable<Model.ApprovableOrganisation>> useCase) =>
+                await useCase.Execute((type, limit, skip))
```
jakzal, line 27: Given that there's a scope for more query parameters added in future, I think we'd be better off with a dedicated query record here, instead of a tupple, i.e. `OrganisationsQuery`. This way it will be easier to introduce additional query params later.

Services/CO.CDP.Organisation.WebApi/Model/ApprovableOrganisation.cs
```diff
@@ -0,0 +1,16 @@
+using CO.CDP.OrganisationInformation;
+
+namespace CO.CDP.Organisation.WebApi.Model;
+public record ApprovableOrganisation
```
jakzal, line 4: The endpoint we should be providing (GET /organisations) is more general purpose. I wouldn't couple this response name to "approvable" organisations only.

Services/CO.CDP.Organisation.WebApi/UseCase/GetApprovableOrganisationsUseCase.cs
```diff
@@ -0,0 +1,61 @@
...
+    }
+    private static string GetRole(OrganisationInformation.Persistence.Organisation organisation)
+    {
+        if (organisation.SupplierInfo != null)
+        {
+            return "supplier";
+        } else if (organisation.BuyerInfo != null)
+        {
+            return "buyer";
+        }
```
jakzal, line 36: This is a wrong way to identify buyers and suppliers.

`Organisation.Roles` tells you what roles organisation plays. It can be more than one. An organisation can be both a buyer and a supplier.

Services/CO.CDP.Organisation.WebApi/Api/Organisation.cs
rmohammed-goaco: I thought it wasn't suitable for this use case since its locked down to a specific user and their access to organisations. Also, this endpoint did not include any pagination properties and thought it high risk to update the existing endpoint.

Services/CO.CDP.Organisation.WebApi/Api/Organisation.cs
rmohammed-goaco: Ok. I'll get this changed.

Services/CO.CDP.Organisation.WebApi/Api/Organisation.cs
rmohammed-goaco: It's still the same (requires a userUrn) ... `FindByUserUrn`
image

I cannot use this endpoint

Services/CO.CDP.Organisation.WebApi/Model/ApprovableOrganisation.cs
rmohammed-goaco: I don't agree. The endpoint is specific for admin access and requires the extra fields

Services/CO.CDP.Organisation.WebApi/UseCase/GetApprovableOrganisationsUseCase.cs
rmohammed-goaco: We discussed this briefly and I still found a mismatch in the DB https://goacoglobal.slack.com/archives/C06M34UGL59/p1726239795145109

jakzal (approved):

PR #641, opened 2024-09-22
Part 3: Data Sharing API endpoints authorization

dharmverma:
Part 3: Data Sharing API endpoints authorization

### DIFF
--- a/Libraries/CO.CDP.Authentication.Tests/Authorization/OrganisationScopeAuthorizationHandlerTests.cs
+++ b/Libraries/CO.CDP.Authentication.Tests/Authorization/OrganisationScopeAuthorizationHandlerTests.cs
@@ -13,15 +13,13 @@ namespace CO.CDP.Authentication.Tests.Authorization;
 
 public class OrganisationScopeAuthorizationHandlerTests
 {
-    private readonly Mock<IHttpContextAccessor> _mockHttpContextAccessor;
-    private readonly Mock<ITenantRepository> _mockTenantRepository;
+    private readonly Mock<IHttpContextAccessor> _mockHttpContextAccessor = new();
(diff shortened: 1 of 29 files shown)


Libraries/CO.CDP.Authentication.Tests/Authorization/OrganisationScopeAuthorizationHandlerTests.cs
```diff
@@ -71,8 +69,8 @@ public async Task HandleRequirementAsync_MatchExpectedResult(bool organisationSc
         var context = CreateAuthorizationHandlerContext(userUrn, organisationScopeExists ? ["Admin"] : ["Scope1"], location);
         MockHttpContext(location, string.Format(urlFormat, organisationId));
 
-        _mockTenantRepository.Setup(x => x.LookupTenant(userUrn))
-            .ReturnsAsync(GetTenantLookup(userUrn, organisationId));
+        _mockOrganisationRepository.Setup(x => x.FindOrganisationPerson(organisationId, userUrn))
+            .ReturnsAsync(GetOrganisationPerson());
```
jakzal, line 73: I'm not sure if that's right. We're suppose to use the tenant lookup for any authorisation requests.

dharmverma: The lookup endpoint is going to be very expensive as it builds complex query and brings lots of information which is not required in authorisation scope check, I have created another version of lookup which creates a nominal query.

dharmverma: Again this is not exposed as endpoint but an internal repo method, lets have a call to discuss why you think its not right?

dpatel017 (approved):

PR #652, opened 2024-09-25
DP-593 Registration - Buyer Approval - Approval Screen

rmohammed-goaco:
https://noticingsystem.atlassian.net/browse/DP-593


Services/CO.CDP.Organisation.WebApi/Api/Organisation.cs
```diff
@@ -47,6 +47,28 @@ public static void UseOrganisationEndpoints(this WebApplication app)
                 return operation;
             });
 
+        app.MapPost("/admin/organisation/review", async (ReviewOrganisation command, IUseCase<ReviewOrganisation, Boolean> useCase) =>
```
rmohammed-goaco, line 50: @jakzal What do you think this endpoint should be? Is it ok how it is or would you prefer something else?

Services/CO.CDP.Organisation.WebApi/UseCase/ReviewOrganisationUseCase.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    public async Task<Boolean> Execute(ReviewOrganisation command)
+    {
+        var organisation = await organisationRepository.Find(command.OrganisationId) ?? throw new UnknownOrganisationException($"Unknown organisation {command.OrganisationId}.");
+        var person = await personRepository.Find(command.approvedById) ?? throw new UnknownPersonException($"Unknown person {command.approvedById}.");
+
+        // Organisation is left in a pending state until user approves it.
+        if (command.Approved)
+        {
+            organisation.ApprovedOn = DateTimeOffset.UtcNow;
```
rmohammed-goaco, line 16: @jakzal At the moment this `ApprovedOn` is what marks the buyer as approved. After the buyer is approved what is it that they will be able to do? Is there somewhere where I need to apply a check that the buyer is approved before actions or something else can be done, or is this functionality to be added at a later date?

Services/CO.CDP.Organisation.WebApi/Api/Organisation.cs
jakzal: Imo we should follow the same pattern as everywhere else. We're working with an organisation resource, so it should be under `/organisations/{organisationId}/...`.

Using `PATCH /organisations/{organisationId}` for any organisation updates would be consistent with the approach the team has used in the past.

Services/CO.CDP.Organisation.WebApi/UseCase/ReviewOrganisationUseCase.cs
jakzal: This is a question for BAs. We need a ticket for this in the backlog imo. What should the buyer be able to do once they're approved? What shouldn't they be able to do before they're approved? Do we return them with a buyer role from our APIs before they're approved?

Services/CO.CDP.Organisation.WebApi/Api/Organisation.cs
rmohammed-goaco: So in this case

PATCH /organisations/{organisationId}/review

Since its not just a simple CRUD task

rmohammed-goaco:
- It is now part of its own group
- The endpoint is now PATCH /support/organisations/{organisationId}
- It uses an enum of types for example: SupportOrganisationUpdateType.Review
- The endpoint is more generic and can be used to update more info down the line

rmohammed-goaco:
Some unit tests still need adding, but wanted feedback on approach first

Frontend/CO.CDP.OrganisationApp/Pages/Support/OrganisationApproval.cshtml.cs
```diff
@@ -0,0 +1,47 @@
...
+namespace CO.CDP.OrganisationApp.Pages.Support;
+
+public class OrganisationApprovalModel(
+    IOrganisationClient organisationClient,
+    ISession session) : LoggedInUserAwareModel(session)
+{
+    public OrganisationWebApiClient.Organisation? OrganisationDetails { get; set; }
+
+    [BindProperty]
+    public Boolean? Approval { get; set; }
```
dharmverma, line 14: Seems like no validation, if user has not selected the option.
Ideally user must select an option before continuing with the page?

dharmverma, line 14: use bool rather Boolean, so we are in sync how we used in other places (common standard across code)

Frontend/CO.CDP.OrganisationApp/Pages/Support/OrganisationApproval.cshtml.cs
```diff
@@ -0,0 +1,47 @@
...
+        {
+            return Redirect("/page-not-found");
+        }
+    }
+
+    public async Task<IActionResult> OnPost(Guid organisationId)
+    {
+        SupportOrganisationInfo orgInfo = new SupportOrganisationInfo(
+            Approval ?? false,
+            UserDetails.PersonId ?? Guid.Empty,
```
dharmverma, line 36: Shouldn't we raise exception if PersonId is null?

Services/CO.CDP.Organisation.WebApi/UseCase/SupportUpdateOrganisationUseCase.cs
```diff
@@ -0,0 +1,35 @@
...
+                var personId = command.supportUpdateOrganisation.Organisation.ApprovedById;
+
+                var person = await personRepository.Find(personId) ?? throw new UnknownPersonException($"Unknown person {personId}.");
+
+                if (command.supportUpdateOrganisation.Organisation.Approved)
+                {
+                    organisation.ApprovedOn = DateTimeOffset.UtcNow;
+                }
+
+                organisation.ApprovedBy = person;
```
dharmverma, line 25: Isn't ApprovedBy is also set only when Approved is set to TRUE

Services/CO.CDP.Organisation.WebApi/UseCase/SupportUpdateOrganisationUseCase.cs
```diff
@@ -0,0 +1,35 @@
...
+                organisation.ApprovedComment = command.supportUpdateOrganisation.Organisation.Comment;
+
+                break;
+        }
+
+        organisationRepository.Save(organisation);
+
+        return await Task.FromResult(true);
```
dharmverma, line 33: You can simply replace this line with  `return true;`

Services/CO.CDP.Organisation.WebApi/Api/Organisation.cs
```diff
@@ -179,6 +179,35 @@ await useCase.Execute(new OrganisationQuery(name, identifier))
         return app;
     }
 
+    public static RouteGroupBuilder UseSupportEndpoints(this RouteGroupBuilder app)
+    {
+        app.MapPatch("/organisation/{organisationId}",
+                [OrganisationAuthorize([AuthenticationChannel.OneLogin])]
```
dharmverma, line 185: Missing Authorization test

rmohammed-goaco: @dharmverma I wanted to ask you about this section actually. I assume there will also be a different role here. What will it be? Still a bit confused. The migration to add Scopes to the person table has been merged into main.

Frontend/CO.CDP.OrganisationApp/Pages/Support/OrganisationApproval.cshtml.cs
rmohammed-goaco: This has been done

Services/CO.CDP.Organisation.WebApi/Api/Organisation.cs
dharmverma: @rmohammed-goaco I am talking about test here.
We have a ticket on board to cover the scope.
All you need to write a test case that only OneLogin users are able to access it.

Services/CO.CDP.Organisation.WebApi/UseCase/SupportUpdateOrganisationUseCase.cs
dharmverma: In that case definitely change the property name, its misleading otherwise.

Frontend/CO.CDP.OrganisationApp/Pages/Support/OrganisationApproval.cshtml.cs
dharmverma: So basically Luke has confirmed that we need to have validation, otherwise how you keep in Pending State on Continue button click?

Services/CO.CDP.Organisation.WebApi/Api/Organisation.cs
rmohammed-goaco: Test case added: https://github.com/cabinetoffice/GCGS-Central-Digital-Platform/pull/652/files#diff-dfa6734ca97b6e1fc83fb71473e1995bd0e27226adf33805fd639a4a321e481e

dharmverma:
Looks good, there are some git warnings, please remove them.

dharmverma (approved):

Services/CO.CDP.Organisation.WebApi/Api/Organisation.cs
```diff
@@ -179,6 +179,35 @@ await useCase.Execute(new OrganisationQuery(name, identifier))
...
+                async (Guid organisationId, SupportUpdateOrganisation supportUpdateOrganisation, IUseCase<(Guid, SupportUpdateOrganisation), bool> useCase) =>
+                await useCase.Execute((organisationId, supportUpdateOrganisation))
+                    .AndThen(Results.Ok))
+            .Produces<Boolean>(StatusCodes.Status200OK, "application/json")
+            .ProducesProblem(StatusCodes.Status500InternalServerError)
+            .ProducesProblem(StatusCodes.Status400BadRequest)
+            .Produces<ProblemDetails>(StatusCodes.Status401Unauthorized)
+            .ProducesProblem(StatusCodes.Status422UnprocessableEntity)
+            .ProducesProblem(StatusCodes.Status404NotFound)
```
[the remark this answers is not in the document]
rmohammed-goaco: Done

dpatel017 (approved):

shilpigoel-goaco (approved):

PR #661, opened 2024-09-25
DP-193 - Identifier Uri mapping added

maciej-goaco:


Services/CO.CDP.OrganisationInformation/CO.CDP.OrganisationInformation.csproj
```diff
@@ -8,4 +8,8 @@
         <NoWarn>$(NoWarn);1591</NoWarn>
     </PropertyGroup>
 
+    <ItemGroup>
+      <PackageReference Include="Microsoft.Extensions.Configuration.Abstractions" Version="8.0.0" />
+    </ItemGroup>
```
jakzal, line 13: Why is it needed here?

maciej-goaco: it's not, removed.

Services/CO.CDP.Organisation.WebApi/IIdentifierService.cs
```diff
@@ -0,0 +1,6 @@
+namespace CO.CDP.Organisation.WebApi;
+
+public interface IIdentifierService
+{
+    string? GetRegistryUri(string? scheme, string? identifierId);
```
jakzal, line 5: Why are `scheme` and `identifierId` nullable?

Services/CO.CDP.Organisation.WebApi/appsettings.json
```diff
@@ -8,6 +8,7 @@
       }
     }
   },
+  "OrganisationApiHostUrl": "http://localhost:8082",
```
jakzal, line 11: "HostUrl" doesn't make much sense - it's either one or the other. Can we name the option `OrganisationApiUrl` or `OrganisationWebApiUrl` to be consistent with `OrganisationAppUrl` we already have?

Services/CO.CDP.Organisation.WebApi/IIdentifierService.cs
maciej-goaco: Scheme previously was and not anymore with current usage, so changed that.
IdentifierId seems to be made purposely nullable, even though not sure what value such entry would bring.

Services/CO.CDP.Organisation.WebApi/appsettings.json
maciej-goaco: done

rmohammed-goaco (approved):

PR #672, opened 2024-09-27
Added optional option to form date and file upload control

dharmverma:
Added optional option to form date and file upload control


dpatel017 (approved):

shilpigoel-goaco (approved):

Frontend/CO.CDP.OrganisationApp/Pages/Forms/FormElementDateInputModel.cs
```diff
@@ -42,11 +50,34 @@ public override void SetAnswer(FormAnswer? answer)
...
+        else if (RedirectFromCheckYourAnswerPage && IsRequired == false)
+        {
+            HasValue = false;
+        }
     }
 
     public IEnumerable<ValidationResult> Validate(ValidationContext validationContext)
     {
-        if (CurrentFormQuestionType == FormQuestionType.Date && IsRequired == true && string.IsNullOrWhiteSpace(DateString))
+        if (CurrentFormQuestionType != FormQuestionType.Date)
```
andymantell, line 61: Why in the DateInputModel would we reach the situation where CurrentFormQuestionType is not equal to Date?

dharmverma: Good point, it may be obsolete now, we had a situation where all forms partial were executing validations.
I will check in my next changes and do clean up

PR #675, opened 2024-09-27
Exclusions - Adding an exclusion - authority website

dharmverma:
Exclusions - Adding an exclusion - authority website


dpatel017 (approved):

Frontend/CO.CDP.OrganisationApp/Pages/Forms/_FormElementUrlInput.cshtml
```diff
@@ -0,0 +1,88 @@
...
+        </p>
+    }
+
+    @if (Model.IsRequired)
+    {
+        RenderField();
+    }
+    else
+    {
+        <fieldset class="govuk-fieldset">
```
andymantell, line 62: Fieldsets must have legends. The pattern you've got is correct for the text input only version. But when you've got this radio field combined with the conditional input field, the h1 and legend would need to come inside the fieldset, and the text field itself would have a label separately from the h1 (It could be visually hidden if necessary for the designs)

andymantell (changes requested):

Services/CO.CDP.OrganisationInformation.Persistence/Forms/FormQuestion.cs
```diff
@@ -35,7 +35,8 @@ public enum FormQuestionType
     Date,
     CheckBox,
     Address,
-    MultiLine
+    MultiLine,
+    Url
```
jakzal, line 39: Do we also need to update the mapping for the data sharing endpoint?

https://github.com/cabinetoffice/GCGS-Central-Digital-Platform/blob/3ed146f0b21806f47aea81fdbdfff45affbbf779/Services/CO.CDP.DataSharing.WebApi/AutoMapper/DataSharingProfile.cs#L122-L169

Frontend/CO.CDP.OrganisationApp/Pages/Forms/_FormElementUrlInput.cshtml
```diff
@@ -0,0 +1,97 @@
...
+    {
+        RenderHeading();
+        RenderField();
+    }
+    else
+    {
+        <fieldset class="govuk-fieldset">
+            <legend class="govuk-fieldset__legend govuk-fieldset__legend--l">
+                @{
+                    RenderHeading();
```
andymantell, line 69: Closer, but still not quite right. The RenderHeading method has too much in it. The legend should be this and nothing more:

```
<legend class="govuk-fieldset__legend govuk-fieldset__legend--l">
    <h1 class="govuk-fieldset__heading">
      @Model.Heading
    </h1>
  </legend>

  ... descriptions, errors and everything else goes below ...
```

I feel like in an attempt to keep this DRY, it's steering away from the right markup in each instance and it would be better to repeat yourself slightly.

andymantell (changes requested):

Frontend/CO.CDP.OrganisationApp/Pages/Forms/_FormElementUrlInput.cshtml
dharmverma: Very well pointed @andymantell

andymantell (approved):

PR #680, opened 2024-09-30
Fix file re-upload validation

dharmverma:
Fix file re-upload validation


mareg (commented):
Should we validate the content type instead of extension?

It's very easy to upload a file with an incorrect file extension, they should not be relied on. It's safer to check for the supported mime content type.

E.g. for the PDF file that would be `application/pdf`.

dbgoaco (approved):

dharmverma:
> Should we validate the content type instead of extension?
> 
> It's very easy to upload a file with an incorrect file extension, they should not be relied on. It's safer to check for the supported mime content type.
> 
> E.g. for the PDF file that would be `application/pdf`.

Yes, we need a proper file check validation that can extract real mimetype by reading file, 
rather extracting mimetype from file extension.
This has been discussed before by @jakzal and I believe we have ticket on board which also should cover virus check.

dpatel017 (approved):

PR #683, opened 2024-10-01
Feature/dp 543 exclusions single choice question type

JBaigGoaco:
The grouped multi choice has been added 

image


andymantell (dismissed):

Frontend/CO.CDP.OrganisationApp/Models/DynamicForms.cs
```diff
@@ -78,5 +79,6 @@ public enum FormQuestionType
     Date,
     CheckBox,
     Address,
-    MultiLine
+    MultiLine,
+    GroupedSingleChoice
```
jakzal, line 83: Do we also need to update the data sharing response with the new type?

https://github.com/cabinetoffice/GCGS-Central-Digital-Platform/blob/3ed146f0b21806f47aea81fdbdfff45affbbf779/Services/CO.CDP.DataSharing.WebApi/AutoMapper/DataSharingProfile.cs#L122-L169

JBaigGoaco: @jakzal Hi Kuba, I think yes we need to return because it will hold at least 5 lines of text messages, i.e. more characters than a normal text box. 
And now we will have Single choice and the GroupedSingleChoice; they will both hold one value.

dpatel017 (approved):

Frontend/CO.CDP.OrganisationApp/Pages/Forms/_FormElementGroupedSingleChoice.cshtml
```diff
@@ -0,0 +1,85 @@
...
+                    {
+                        foreach (var choice in group.Choices)
+                        {
+                            var id = $"SelectedOption_{choice.Value}";
+
+                            <div class="govuk-radios__item">
+                                <input class="govuk-radios__input" id="@id" name="SelectedOption" type="radio" value="@choice.Value" @(Model.SelectedOption == choice.Value ? "checked" : "")>
+                                <label class="govuk-label govuk-radios__label" for="@id">@choice.Title</label>
+                            </div>
+                        }
```
andymantell, line 77: The ID for the first radio button needs to be "SelectedOption", so that it matches up with the in page anchor link in the error summary when validation fails. This means that when you fail validation and click the link, focus is moved to the first radio button in the set. What I normally do is make them numerical and omit the number for the first one.

```suggestion
                    @if (group.Choices != null)
                    {
                        var index = 0;
                        foreach (var choice in group.Choices)
                        {
                            var id = index == 0 ? "SelectedOption" : $"SelectedOption_{index}";

                            <div class="govuk-radios__item">
                                <input class="govuk-radios__input" id="@id" name="SelectedOption" type="radio" value="@choice.Value" @(Model.SelectedOption == choice.Value ? "checked" : "")>
                                <label class="govuk-label govuk-radios__label" for="@id">@choice.Title</label>
                            </div>
                            
                            index++;
                        }
```

andymantell (changes requested):

Frontend/CO.CDP.OrganisationApp/Pages/Forms/_FormElementGroupedSingleChoice.cshtml
JBaigGoaco: @andymantell Thats a really good point Andy. Ill commit that suggestion now.

andymantell (approved):

PR #721, opened 2024-10-04
Prevent admin from succeeding support admin requirement by default.

rmohammed-goaco:
The Support nav was showing when entering an organisation context. This was because Admin was supposed to override Editor and Viewer scopes and allow them to do anything within an org. 

We forgot to check the requirement wasn't for a personScope though, specifically SUPPORTADMIN. This takes care of that.


Frontend/CO.CDP.OrganisationApp/Authorization/CustomScopeHandler.cs
```diff
@@ -23,7 +23,7 @@ protected override async Task HandleRequirementAsync(AuthorizationHandlerContext
                     var organisationUserScopes = await _userInfo.GetOrganisationUserScopes();
 
                     // Admin role can do anything within this organisation
-                    if (organisationUserScopes.Contains(OrganisationPersonScopes.Admin))
+                    if (organisationUserScopes.Contains(OrganisationPersonScopes.Admin) && requirement.Scope != PersonScopes.SupportAdmin)
```
dharmverma, line 26: Doesn't seems right, if the person is admin for organisation he should be able to perform anything for that organisation, regardless of person scope

rmohammed-goaco: Yeah, you're right. This does not cover that scenario.

rmohammed-goaco: Wouldn't the next pass of requirement.Scope equals ADMIN, EDITOR or VIEWER satisfy this condition?

andymantell: This isn't testing the person's scope though, it's checking that the _requirement_ that we're testing for is not support admin.

If the person is admin, but the thing we're looking for is support admin, don't let them through.

andymantell (approved):

Frontend/CO.CDP.OrganisationApp/Authorization/CustomScopeHandler.cs
dharmverma: @andymantell  what if support admin has a "test organisation" to do checks?
Also if that is not possible scenario, than also we are sure that Person with 'SupportAdmin' scope will never be and Admin on an Organisation hence the above condition will never met?

andymantell: That doesn't matter - because we're testing the _requirement_, not the presence of support admin on their user. This line is saying, if the _requirement_ is asking for support admin, but the current person is admin, don't allow them through.

andymantell: Wait I see what you're saying I think

rmohammed-goaco: Yeah. I think the line of code is just confusing because its checking for the presence of an organisation scope on the left and then checking that there isn't a SupportAdmin person scope requirement on the right.

I'll refactor to make the logic's intention clearer

dharmverma (approved):

Frontend/CO.CDP.OrganisationApp.Tests/CustomScopeHandlerTests.cs
```diff
@@ -0,0 +1,88 @@
...
+using CO.CDP.OrganisationApp.Constants;
+using Microsoft.AspNetCore.Authorization;
+using Moq;
+using FluentAssertions;
+using Microsoft.Extensions.DependencyInjection;
+using Xunit;
+
+namespace CO.CDP.OrganisationApp.Tests.Authorization;
+
+public class CustomScopeHandlerTests
```
rmohammed-goaco, line 11: @andymantell @dharmverma Added a chunk of test scenarios. Thanks for your help guys.

andymantell: Nice 👍🏻

dharmverma:
some Github warning popping up

andymantell:
> some Github warning popping up

Yeah that's my fault on a previous PR 🤦🏻‍♂️ 
Since I think this is probably the next PR to go into main I've asked Richard if he would be so kind as to clean up after me 😬

PR #742, opened 2024-10-10
DP-626 Request to join org migration

rmohammed-goaco:


Services/CO.CDP.OrganisationInformation/JoinRequestStatus.cs
```diff
@@ -0,0 +1,11 @@
+using System.Text.Json.Serialization;
+
+namespace CO.CDP.OrganisationInformation;
+
+[JsonConverter(typeof(JsonStringEnumConverter))]
+public enum JoinRequestStatus
```
dharmverma, line 6: Suggestion, can we rename to `RequestStatus` ?

Services/CO.CDP.OrganisationInformation.Persistence/JoinRequest.cs
```diff
@@ -0,0 +1,21 @@
+using CO.CDP.EntityFrameworkCore.Timestamps;
+using Microsoft.EntityFrameworkCore;
+
+namespace CO.CDP.OrganisationInformation.Persistence;
+
+[Index(nameof(Guid), IsUnique = true)]
+public class JoinRequest : IEntityDate
```
dharmverma, line 7: Suggestion, table name - OrganisationPersonAddRequest ?

Services/CO.CDP.OrganisationInformation.Persistence/JoinRequest.cs
```diff
@@ -0,0 +1,21 @@
...
+{
+    public int Id { get; set; }
+    public required Guid Guid { get; set; }
+    public required Organisation Organisation { get; set; }
+    public required int OrganisationId { get; set; }
+    public required Person Person { get; set; }
```
dharmverma: As discussed previously, please mark relational property nullable, so that we don't need to forcefully fetch or set when dealing with entity

Services/CO.CDP.OrganisationInformation.Persistence/JoinRequest.cs
rmohammed-goaco: I still prefer join request. Had a session with @jakzal and we thought it appropriate. See https://lucid.app/lucidspark/cb2c1600-0bff-42f2-b5d1-d3c96321b0ca/edit?invitationId=inv_3c51f859-f04a-4885-aff3-a3d949d11acd&page=0_0# for more info

Services/CO.CDP.OrganisationInformation/JoinRequestStatus.cs
rmohammed-goaco: That seems too generic to me. Could mean any kind of request then in another developer's eyes and could be confusing. HttpRequest, for example, also has a status.

rmohammed-goaco: @jakzal thoughts?

Services/CO.CDP.OrganisationInformation.Persistence/JoinRequest.cs
jakzal: Naming it `OrganisationJoinRequest` wouldn't hurt while giving more context?

dharmverma (approved):

dpatel017 (approved):

JBaigGoaco (approved):

jakzal (approved):

PR #748, opened 2024-10-11
Feature/dp 705 vat cannot be primary identifier

dbgoaco:
[DP-705](https://noticingsystem.atlassian.net/browse/DP-705)
- VAT must not be shown on organisation screen;
- VAT must not be a primary identifier;
- Migration removes any VAT primary identifiers and sets the first non-VAT identifier as the primary.


JBaigGoaco (approved):

andymantell (changes requested):
Have DM'd you - I think the migration might not be right but would like to check with you.

andymantell (approved):

PR #749, opened 2024-10-11
DP-644 - fix skip links not working on all pages because main element was missing id attribute

andymantell:


rmohammed-goaco (approved):
You're right, that was quick. I hope you did a find/replace all otherwise that must have been painful.

maciej-goaco (approved):

PR #750, opened 2024-10-11
Make repository tests less likely to fail due to parallel runs

jakzal:
I noticed these tests sometimes fail.

It turns out it's due to one missing `await` call and fixtures shared between test cases.

Testcontainers are only started once per test class. It's a compromise that make tests run faster. The drawback is data created in a previous test case can affect next ones. It's important each test case creates a unique data set. We do have a solution for this (starting a transaction that's rolled back at the end), but we only use it as a last resort.

### DIFF
--- a/Services/CO.CDP.OrganisationInformation.Persistence.Tests/DatabaseAuthenticationKeyRepositoryTest.cs
+++ b/Services/CO.CDP.OrganisationInformation.Persistence.Tests/DatabaseAuthenticationKeyRepositoryTest.cs
@@ -35,12 +35,12 @@ public class DatabaseAuthenticationKeyRepositoryTest(PostgreSqlFixture postgreSq
         using var repository = AuthenticationKeyRepository();
         var organisation = GivenOrganisation();
         var authenticationKey = GivenAuthenticationKey(key: Guid.NewGuid().ToString(), organisation: organisation);
-        authenticationKey.Organisation = GivenOrganisation();
 
         await repository.Save(authenticationKey);
 
         var found = await repository.Find(authenticationKey.Key);
-        found.As<AuthenticationKey>().OrganisationId.Should().Be(authenticationKey.Organisation.Id);
+        found.As<AuthenticationKey>().OrganisationId.Should().NotBeNull();
+        found.As<AuthenticationKey>().OrganisationId.Should().Be(authenticationKey.Organisation?.Id);
     }
 
     [Fact]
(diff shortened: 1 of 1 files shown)


Services/CO.CDP.OrganisationInformation.Persistence.Tests/DatabaseAuthenticationKeyRepositoryTest.cs
```diff
@@ -62,10 +62,10 @@ public async Task ItFindsSavedAuthenticationKeyByNameKeyAndOrganisationId()
         using var repository = AuthenticationKeyRepository();
         var organisation = GivenOrganisation();
         var key = Guid.NewGuid().ToString();
-        var authenticationKey = GivenAuthenticationKey(key: key, organisation: organisation);
+        var authenticationKey = GivenAuthenticationKey(name: "fts-key", key: key, organisation: organisation);
```
jakzal, line 65: Details that are relevant to the test case should be visible in the test and not hidden. This name is used to find the record later, so I made it visible so it's clear where the name comes from.

Services/CO.CDP.OrganisationInformation.Persistence.Tests/DatabaseAuthenticationKeyRepositoryTest.cs
```diff
@@ -76,27 +76,27 @@ public void ItRejectsTwoApiKeyWithTheSameName()
...
         Organisation? organisation = null,
         List<string>? scopes = null
     )
     {
         return new AuthenticationKey
         {
-            Name = name,
-            Key = key,
+            Name = name ?? $"key {Guid.NewGuid()}",
+            Key = key ?? Guid.NewGuid().ToString(),
```
jakzal, line 99: I randomised details, so it's:
*  unlikely two tests create the same entry
* person writing the next test relies on generated details but will pass them in instead

Services/CO.CDP.OrganisationInformation.Persistence.Tests/DatabaseAuthenticationKeyRepositoryTest.cs
```diff
@@ -76,27 +76,27 @@ public void ItRejectsTwoApiKeyWithTheSameName()
...
-        var authenticationKey1 = GivenAuthenticationKey(key: key, organisation: organisation);
-        var authenticationKey2 = GivenAuthenticationKey(key: key, organisation: organisation);
+        var authenticationKey1 = GivenAuthenticationKey(name: "fts", key: key, organisation: organisation);
+        var authenticationKey2 = GivenAuthenticationKey(name: "fts", key: key, organisation: organisation);
 
         repository.Save(authenticationKey1);
-        
-        repository.Invoking(r => r.Save(authenticationKey2))
+
+        repository.Invoking(async r => await r.Save(authenticationKey2))
```
jakzal, line 84: Missing `await` here was making this test occasionally fail.

shilpigoel-goaco (approved):

maciej-goaco (approved):

PR #759, opened 2024-10-15
fixed API error

shilpigoel-goaco:


Frontend/CO.CDP.OrganisationApp/Constants/ApiExceptionMapper.cs
```diff
@@ -0,0 +1,37 @@
...
+
+public static class ApiExceptionMapper
+{
+    public static void MapApiExceptions(
+        ApiException<ProblemDetails> aex,
+        ModelStateDictionary modelState)
+    {
+        var code = ExtractErrorCode(aex);
+
+        if (!string.IsNullOrEmpty(code))
```
jakzal, line 14: What will happen if code is null or empty? I'm guessing no error will be shown. Since this method is called when an exception is thrown, is it what we expect?

JBaigGoaco (approved):

Frontend/CO.CDP.OrganisationApp/Constants/ApiExceptionMapper.cs
shilpigoel-goaco: refactored the code

jakzal (approved):

PR #760, opened 2024-10-15
Feature/dp 670 manipulating urls records creating bug

JBaigGoaco:
Fixed issued in forms where manipulating the URLs records can create new database entries and jumping onto wrong pages in the journey.  

**Details:**
manipulating the URLs records and passing question IDs manually can create new data in the DB and navigates to different pages which is fixed in this PR. when someone will hack the URLs we will bring them back to the valid question type based on the locally stored answersSet and determining the order of the questions based on SortOrder number to go to the valid screen in the Journey.


andymantell (approved):

Services/CO.CDP.Forms.WebApi/Model/FormQuestion.cs
```diff
@@ -15,6 +15,8 @@ public record FormQuestion
     public string? Caption { get; init; }
     public string? SummaryTitle { get; set; }
     public required FormQuestionOptions Options { get; init; } = new();
+    public string? Name { get; set; }
+    public int? SortOrder { get; init; }
```
jakzal, line 19: We're suppose to determine the next question to answer based on the `nextQuestionId`, and not the sort order. Sort order was only introduced to pre-sort questions in responses. For that reason, it's unnecessary to expose the sort order here. I'd even say it's undesired as it might encourage the wrong use.

JBaigGoaco: @jakzal Hi Kuba , the issue or bug was raised regarding people manipulating the URL and passing in the question IDs in the URLs to navigate to the different pages. we stopped that issues and I had to use this Sort Order to find out which question we should redirect to when someone try to hack the URL so SortOrder was needed to bring or Navigate back to the correct Page in the Journey based on Sort Order.

jakzal: The problem is that the sort order doesn't tell you which question is previous or next. The order of questions should be determined by nextQuestion/nextQuestionAlternative. Remember, there are multiple paths you can take answering the questions.

JBaigGoaco: yes sure, originally i started working on the basis of  nextQuestion but it was getting complex and complex but ill revisit and think in that direction again

shilpigoel-goaco (approved):

PR #761, opened 2024-10-16
DP-695 - Error: Failed during call to registered office address.

dbgoaco:
[DP-695](https://noticingsystem.atlassian.net/jira/software/projects/DP/boards/22?selectedIssue=DP-695)

Companies house API error logging only occurs for non 500/404 errors.

Added initial version of logging error codes.


rmohammed-goaco (approved):

JBaigGoaco (approved):

Frontend/CO.CDP.OrganisationApp/ThirdPartyApiClients/CompaniesHouse/CompaniesHouseApi.cs
```diff
@@ -46,14 +51,24 @@ public class CompaniesHouseApi(IConfiguration configuration,
...
-            logger.LogError(exc, "Failed during call to get company profile.");
+            Log(exc, companyNumber);
         }
 
         return profile;
     }
 
+    private void Log(Exception exc, string companyNumber)
+    {
+        var logException = new ExceptionLogging($"Failed to call Companies House API for company number: {companyNumber}.", ErrorCodes.CompaniesHouseApiError, exc);
```
jakzal, line 67: `companyNumber` is assigned an error code, instead of the company number.

Frontend/CO.CDP.OrganisationApp/Logging/ErrorCodes.cs
```diff
@@ -0,0 +1,6 @@
+namespace CO.CDP.OrganisationApp.Logging;
+
+public class ErrorCodes
+{
+    public const int CompaniesHouseApiError = 1000;
```
jakzal, line 5: This needs to be something easily identifiable. `1000` can easily show up in third party log entries of any random library we use.

Imo, error codes should be strings that share common prefix, i.e. `SIRSI1000` or `CDP1000`. We need to decide on the prefix.

ping @mareg

Frontend/CO.CDP.OrganisationApp/ThirdPartyApiClients/CompaniesHouse/CompaniesHouseApi.cs
```diff
@@ -46,14 +51,24 @@ public class CompaniesHouseApi(IConfiguration configuration,
                 .GetAsync()
                 .ReceiveJson<CompanyProfile>();
         }
+        catch (FlurlHttpException ex) when ((ex.StatusCode == (int)HttpStatusCode.NotFound) || (ex.StatusCode == (int)HttpStatusCode.InternalServerError))
+        {
+        }
```
jakzal, line 56: Aren't we swallowing 500 errors this way?

dbgoaco: Yes, but if we don't do this we'll end up with the same errors in the logs whenever an incorrect company number is entered during registration.

Frontend/CO.CDP.OrganisationApp/Logging/ErrorCodes.cs
dbgoaco: Agreed.

Frontend/CO.CDP.OrganisationApp/ThirdPartyApiClients/CompaniesHouse/CompaniesHouseApi.cs
dbgoaco: The second param to ExceptionLogging is the error code, not the company number for the string.

jakzal: The message only has one placeholder, and it's for the company number.

Frontend/CO.CDP.OrganisationApp/ThirdPartyApiClients/CompaniesHouse/CompaniesHouseApi.cs
jakzal: If there's no way to distinguish between the two types of 500 errors (invalid number vs an actual problem with the api), then I think we'll need to log all 500 errors :(

Otherwise, we'll miss the cases when there's an actual problem with the API.

dbgoaco: Yeah, this was the dilemma I faced when I developed it, which is why I thought it wise to validate the company number, but Sammy would prefer us not to do it.

dbgoaco: Ok, I've excluded 404s as these aren't really an error.

jakzal (approved):

PR #767, opened 2024-10-16
API method to get all user with pending request to join organisation and unit test

dpatel017:
User Management - Approve User who requested to join


JBaigGoaco (approved):

Services/CO.CDP.Organisation.WebApi/UseCase/UpdateJoinRequestUseCase.cs
```diff
@@ -0,0 +1,89 @@
...
+        joinRequest.ReviewedOn = DateTimeOffset.UtcNow;
+
+        if (command.updateJoinRequest.status == OrganisationInformation.OrganisationJoinRequestStatus.Accepted)
+        {
+            organisation.OrganisationPersons.Add(new OrganisationPerson
+            {
+                Person = joinRequest.Person!,
+                Organisation = organisation,
+                Scopes = []
+            });
```
rmohammed-goaco, line 39: You'll also need to add a relationship between the person and the company's tenant, otherwise it won't show up on the user's list. Something like: 

```
person.Tenants.Add(organisation.Tenant);
personRepository.Save(person);
```

rmohammed-goaco (changes requested):

Services/CO.CDP.Organisation.WebApi/UseCase/UpdateJoinRequestUseCase.cs
rmohammed-goaco: Actually it would be 
```
joinRequest.Person.Tenants.Add(organisation.Tenant);
personRepository.Save(joinRequest.Person);
```

And you'll also need to add the following line to the OrgRepo to include the tenants
```
.Include(p => p.Tenant)
```

as in

```
    public async Task<Organisation?> Find(Guid organisationId)
    {
        return await context.Organisations
            .Include(p => p.Tenant)
            .Include(p => p.Addresses)
            .ThenInclude(p => p.Address)
            .FirstOrDefaultAsync(t => t.Guid == organisationId);
    }
```

dpatel017: Done

rmohammed-goaco (approved):

PR #794, opened 2024-10-22
DP-753 - Annotations updated for Trusts/Trustees

maciej-goaco:


jakzal (approved):

jakzal:
From PostgreSQL docs:

> An enum value occupies four bytes on disk. The length of an enum value's textual label is limited by the NAMEDATALEN setting compiled into PostgreSQL; in standard builds this means at most 63 bytes.

https://www.postgresql.org/docs/current/datatype-enum.html

dbgoaco (commented):
How did we know what to update the annotations to? They aren't defined in the story.

maciej-goaco:
Replaced with https://github.com/cabinetoffice/GCGS-Central-Digital-Platform/pull/862

PR #807, opened 2024-10-23
DP-731 - Duplicate Checks - Additional Identifier and direct user to request to join

rmohammed-goaco:
https://noticingsystem.atlassian.net/browse/DP-731


Frontend/CO.CDP.OrganisationApp/Pages/Registration/OrganisationAlreadyRegistered.cshtml
```diff
@@ -32,10 +32,9 @@
                     If you know the person who registered the organisation, they can invite you to join. Or you can request to join below.
                 </p>
 
-                <govuk-button>
+                <a class="govuk-button" href="/registration/@Model.Identifier/join-organisation">
```
andymantell, line 35: If we're no longer posting a form with a button here, then the form on line 22 should go as well

Frontend/CO.CDP.OrganisationApp/Pages/Registration/OrganisationAlreadyRegistered.cshtml.cs
```diff
@@ -1,7 +1,16 @@
+using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.RazorPages;
 
 namespace CO.CDP.OrganisationApp.Pages.Registration;
 
 public class OrganisationAlreadyRegistered: PageModel
 {
+    [BindProperty]
```
andymantell, line 8: Bit confused how this page is being used? `[BindProperty]` is for data that is coming from get or post data. The identifier is already sufficiently retrived from the URL segment (`@page "/registration/organisation-already-registered/{Identifier}"`) by the `OnGet(string identifier)` so you don't need the Bind Property at all?

andymantell (changes requested):

Frontend/CO.CDP.OrganisationApp/Pages/Registration/OrganisationAlreadyRegistered.cshtml.cs
```diff
@@ -1,7 +1,16 @@
...
+    public string? Identifier { get; set; }
+
+    public IActionResult OnGet(string identifier)
+    {
+        Identifier = identifier;
```
andymantell, line 13: Bit confused how this page is being used? `[BindProperty]` is for data that is coming from get or post data. The identifier is already sufficiently retrived from the URL segment (`@page "/registration/organisation-already-registered/{Identifier}"`) by the `OnGet(string identifier)` so you don't need the Bind Property at all?

andymantell (changes requested):

Frontend/CO.CDP.OrganisationApp/Pages/Registration/OrganisationAlreadyRegistered.cshtml.cs
rmohammed-goaco: I'll try without and see what happens

Frontend/CO.CDP.OrganisationApp/Pages/Registration/OrganisationAlreadyRegistered.cshtml
rmohammed-goaco: Will remove

andymantell (approved):

dharmverma (approved):

PR #808, opened 2024-10-23
DP-741 - Db Trigger added to automatically update Questions sort order

maciej-goaco:
on table insert and sort_order and next_question_id fields' update


dbgoaco (commented):
Have we considered the effects of using db triggers on Entity Framework?

maciej-goaco:
Replaced with https://github.com/cabinetoffice/GCGS-Central-Digital-Platform/pull/875

PR #809, opened 2024-10-23
Fixed - forms answer no selection value display

dharmverma:
Fixed - forms answer no selection value display


JBaigGoaco (approved):

andymantell (commented):
So this only saves data when HasValue is true. But will it clear out the old data if you had already clicked yes in a previous journey? And then go through to change and then click no?

dharmverma:
> So this only saves data when HasValue is true. But will it clear out the old data if you had already clicked yes in a previous journey? And then go through to change and then click no?

It only save data when HasValue is null or true, and also clears the value if editing again an option selected is NO

andymantell (approved):

PR #819, opened 2024-10-24
DP-723: Configure data protection to persist keys to AWS System Manager

maciej-goaco:


maciej-goaco:
Created for test purpose only

PR #826, opened 2024-10-24
Bugfix/dp 765 check person not already part of org

rmohammed-goaco:


Frontend/CO.CDP.OrganisationApp/Pages/Registration/JoinOrganisation.cshtml.cs
```diff
@@ -1,20 +1,25 @@
...
+    ITempDataService tempDataService) : LoggedInUserAwareModel(session)
 {
     public OrganisationWebApiClient.Organisation? OrganisationDetails { get; set; }
 
     [BindProperty]
     [Required(ErrorMessage = "Select an option")]
     public bool Join { get; set; }
 
+    public FlashMessage NotificationBannerAlreadyMemberOfOrganisation { get { return new FlashMessage("You are already a member of this organisation. <a class='govuk-notification-banner__link' href='/organisation/" + OrganisationDetails?.Id + "'>View Organisation</a>"); } }
+
```
shilpigoel-goaco, line 22: One question, Can't message defined in ErrorMessageList be used here?

rmohammed-goaco: Yeah. I'll put that in

andymantell (approved):

Services/CO.CDP.Organisation.WebApi/UseCase/CreateOrganisationJoinRequestUseCase.cs
```diff
@@ -42,6 +42,14 @@ public async Task<OrganisationJoinRequest> Execute((Guid organisationId, CreateO
         var person = await personRepository.Find(command.createOrganisationJoinRequestCommand.PersonId)
                            ?? throw new UnknownPersonException($"Unknown person {command.createOrganisationJoinRequestCommand.PersonId}.");
 
+        var alreadyAdded = await CheckPersonAlreadyAdded(organisation, person);
+
+        if (alreadyAdded)
+        {
+            throw new PersonAlreadyAddedToOrganisationException(
+                $"Person {command.createOrganisationJoinRequestCommand.PersonId} already added to organisation {command.organisationId}.");
+        }
```
jakzal, line 51: All these added lines seem to belong together. They accomplish one thing - guard that the person isn't already added. Why not replace all these lines with a single `async GuardPersonIsNotAlreadyAdded(organisation, person)`?

`CheckPersonAlreadyAdded` could be just renamed to `GuardPersonIsNotAlreadyAdded` and throw exception instead of returning a bool. 

Outcome: less code, and code in the public method remains on the same level of abstraction.

dbgoaco (commented):
Works but error message doesn't indicate that it's a duplicate:
image

dbgoaco (commented):
Doesn't work for me. I get the following error, when I go back to the users screen I see duplicate still:
image

rmohammed-goaco:
@dbgoaco Did you rebuild everything?

Services/CO.CDP.Organisation.WebApi/UseCase/CreateOrganisationJoinRequestUseCase.cs
rmohammed-goaco: @jakzal Thanks for the feedback. Done 👍

maciej-goaco (approved):

PR #832, opened 2024-10-25
DP-617 - Claiming a person invite as the user that created the organisation causes an error

maciej-goaco:


maciej-goaco:
Moved the implementation to https://github.com/cabinetoffice/GCGS-Central-Digital-Platform/pull/833 as better fits the new, dedicated ticket https://noticingsystem.atlassian.net/browse/DP-773

PR #848, opened 2024-10-25
Change for user friendly message for the duplicate identifier number

shilpigoel-goaco:


Frontend/CO.CDP.OrganisationApp/Constants/ApiExceptionMapper.cs
```diff
@@ -42,6 +44,7 @@ private static string GetErrorMessageByCode(string errorCode)
             ErrorCodes.UNKNOWN_ORGANISATION => ErrorMessagesList.UnknownOrganisation,
             ErrorCodes.BUYER_INFO_NOT_EXISTS => ErrorMessagesList.BuyerInfoNotExists,
             ErrorCodes.UNKNOWN_BUYER_INFORMATION_UPDATE_TYPE => ErrorMessagesList.UnknownBuyerInformationUpdateType,
+            ErrorCodes.ORGANISATION_UPDATE_INVALID_INPUT => exceptionMessage ?? ErrorMessagesList.UnexpectedError,
```
jakzal, line 47: Do I understand this correctly, that this way we'll expose the message returned by the API?

If that's the case, we initially disregarded this approach and decided APIs should return error codes that are mapped to a friendly message by the frontend. What has changed?

jakzal (changes requested):

Frontend/CO.CDP.OrganisationApp/Constants/ApiExceptionMapper.cs
jakzal: My concern is about using the exception message in the frontend at all. We are not supposed to do that according to the approach with error codes we agreed to.

So let me repeat the question: Do I understand this correctly, that this way we'll expose the message returned by the API?

Using null forgiving is definitely not the right solution here. Do we even expose the `Detail` in `Production`?

shilpigoel-goaco: message coming from the API will be exposed to frontend especially for "InvalidUpdateOrganisationCommand" exception class. As this class is used to throw multiple messages based on the validation, mostly Invalid input.  
It is very difficult to map one exception with multiple message at front end.  and I couldn't find a better way to show user friendly message unless if we just want to show "Invalid Input".

jakzal: We agreed we should not expose the api messages. Error codes should be mapped to messages by the frontend. The Detail you reference is not even populated in Production. You need a more specific error code returned by the API.

shilpigoel-goaco: It will be at least 10 new exception classes added, let me ask business if they are happy with generic message "Invalid Input"

jakzal: This has much bigger implications. Messages need to be translated. We do not translate API messages, and it would have a huge impact if we wanted to start.

jakzal: Not even mentioning the API isn't helpful at all by returning a generic error code in all validation cases.

By the way, you do not need a new exception class for each code (although it has its benefits).

If you wanted to avoid it, you could provide factories for exceptions:

```csharp
 throw InvalidUpdateOrganisationCommand.MissingIdentifierNumberForScheme(identifier.Scheme);
```

The exception would be created with code:

```csharp
public class InvalidUpdateOrganisationCommand(string message, string code = "INVALID_UPDATE_ORGANISATION_COMMAND", Exception? cause = null) : Exception(message, cause)
{
    public static InvalidUpdateOrganisationCommand MissingIdentifierNumberForScheme(string scheme)
    {
        return new InvalidUpdateOrganisationCommand($"Missing Identifier Number for scheme '{scheme}'", "MISSING_IDENTIFIER_NUMBER");
    }
}
```

The code then can be returned from the API. 

I think it's important that the code is encapsulated rather than passed from the outside.


However, not sure if it's much better than:

```csharp
 throw new InvalidUpdateOrganisationCommand.MissingIdentifierNumberForScheme(identifier.Scheme);
```

```csharp
public class InvalidUpdateOrganisationCommand(string message, Exception? cause = null) : Exception(message, cause)
{
    public class MissingIdentifierNumberForScheme(string scheme)
        : InvalidUpdateOrganisationCommand($"Missing Identifier Number for scheme '{scheme}'.");
}
```

Services/CO.CDP.Organisation.WebApi.Tests/UseCase/UpdateOrganisationUseCaseTest.cs
```diff
@@ -150,6 +151,77 @@ await act.Should()
...
+
+        _organisationRepositoryMock.Setup(repo => repo.FindByIdentifier("VAT", "13294342")).ReturnsAsync(organisation);
+
+
+
+        Func<Task> act = async () => await UseCase.Execute((_anotherOrganisationId, command));
+
+
+        await act.Should()
+          .ThrowAsync<InvalidUpdateOrganisationCommand>()
```
jakzal, line 184: This should be a more specific exception. 

Then, there's no need to check the exception message if we already verify the type.

Services/CO.CDP.Organisation.WebApi.Tests/UseCase/UpdateOrganisationUseCaseTest.cs
```diff
@@ -311,10 +383,10 @@ public async Task Execute_ShouldNotInsertIdentifier_WhenIdentifierIdIsEmptyOrNul
         var organisation = OrganisationWithPponIdentifier;
         _organisationRepositoryMock.Setup(repo => repo.Find(_organisationId)).ReturnsAsync(organisation);
 
-        var result = await UseCase.Execute((_organisationId, command));
+        Func<Task> act = async () => await UseCase.Execute((_organisationId, command));
 
-        result.Should().BeTrue();
-        _organisationRepositoryMock.Verify(repo => repo.SaveAsync(organisation, AnyOnSave()), Times.Once);
+        await act.Should()
+           .ThrowAsync<InvalidUpdateOrganisationCommand>();
```
jakzal, line 389: Expect a more specific exception please.

Services/CO.CDP.Organisation.WebApi.Tests/UseCase/UpdateOrganisationUseCaseTest.cs
```diff
@@ -628,6 +700,37 @@ await act.Should()
             Roles = [PartyRole.Buyer]
         };
 
+    private Persistence.Organisation AnotherOrganisation =>
```
jakzal, line 703: Can we name this so it reflects what's different/important about this organisation?

Services/CO.CDP.Organisation.WebApi.Tests/UseCase/UpdateOrganisationUseCaseTest.cs
shilpigoel-goaco: well spotted.

Done.

Services/CO.CDP.Organisation.WebApi/UseCase/UpdateOrganisationUseCase.cs
```diff
@@ -97,23 +97,32 @@ public async Task<bool> Execute((Guid organisationId, UpdateOrganisation updateO
...
 
                 foreach (var identifier in updateObject.AdditionalIdentifiers)
                 {
+                    if (string.IsNullOrWhiteSpace(identifier.Id))
+                    {
+                        throw new InvalidUpdateOrganisationCommand.MissingIdentifierNumber();
+                    }
+
+                    // Check if Identifier number already exists
+                    var organisationIdentifier = organisationRepository.FindByIdentifier(identifier.Scheme, identifier.Id);
```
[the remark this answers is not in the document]
shilpigoel-goaco: done

Services/CO.CDP.Organisation.WebApi.Tests/UseCase/UpdateOrganisationUseCaseTest.cs
shilpigoel-goaco: It is just AnotherOrganisation, nothing specific. when user try to add additional identifiers with already existing identifier number then it creates issue.

jakzal: That's why I prefer factory methods. What are you going to when a third variant of organisation is needed?

Services/CO.CDP.Organisation.WebApi.Tests/UseCase/UpdateOrganisationUseCaseTest.cs
```diff
@@ -311,16 +381,16 @@ public async Task Execute_ShouldNotInsertIdentifier_WhenIdentifierIdIsEmptyOrNul
         var organisation = OrganisationWithPponIdentifier;
         _organisationRepositoryMock.Setup(repo => repo.Find(_organisationId)).ReturnsAsync(organisation);
 
-        var result = await UseCase.Execute((_organisationId, command));
+        Func<Task> act = async () => await UseCase.Execute((_organisationId, command));
 
-        result.Should().BeTrue();
-        _organisationRepositoryMock.Verify(repo => repo.SaveAsync(organisation, AnyOnSave()), Times.Once);
+        await act.Should()
+           .ThrowAsync<InvalidUpdateOrganisationCommand>();
```
jakzal, line 387: This is still not a specific exception.

dharmverma (approved):

Services/CO.CDP.Organisation.WebApi/UseCase/UpdateOrganisationUseCase.cs
```diff
@@ -97,23 +97,22 @@ public async Task<bool> Execute((Guid organisationId, UpdateOrganisation updateO
...
             case OrganisationUpdateType.AdditionalIdentifiers:
-                if (updateObject.AdditionalIdentifiers == null)
+                if (updateObject.AdditionalIdentifiers == null || !updateObject.AdditionalIdentifiers.Any())
                 {
                     throw new InvalidUpdateOrganisationCommand.MissingAdditionalIdentifiers();
                 }
 
                 foreach (var identifier in updateObject.AdditionalIdentifiers)
                 {
+                    await ValidateIdentifierIsNotKnownToUs(organisationRepository, identifier);
```
jakzal, line 107: No need to pass the `organisationRepository` here, as it's part of the class properties.

Services/CO.CDP.Organisation.WebApi/UseCase/UpdateOrganisationUseCase.cs
```diff
@@ -195,6 +194,20 @@ await organisationRepository.SaveAsync(organisation,
         return await Task.FromResult(true);
     }
 
+    private static async Task ValidateIdentifierIsNotKnownToUs(IOrganisationRepository organisationRepository, OrganisationIdentifier identifier)
+    {
+        if (string.IsNullOrWhiteSpace(identifier.Id))
```
jakzal, line 199: Are you sure about this? Just double-checking as I'm not sure right now if this use case is used to persist the "Other" identifier that has the `Id` set to `null`. Do we want to disallow "Other" identifiers here?

shilpigoel-goaco: there is a separate ticket in the backlog to update additionalIdentifiers so it will be dealt there.

jakzal: I'm not sure if I understand your response.

We didn't use to throw an exception if the identifier id is empty. Now we do. Did you make sure this change didn't break adding/updating an "Other" identifier that has the Id empty?

shilpigoel-goaco: user cant add 'Other' identifier on the update organisation in current scenario so it is fine for now. But when Update Identifier ticket will come into picture then It might require removing this check but I dont know use case for that.
If you see that this check is not required then I can remove it?

jakzal: No, leave it in. I just needed to make sure you checked your bases.

jakzal (approved):

PR #849, opened 2024-10-25
Dp 747 again missed updates

jason-goaco:
Just reviewing this DP-747 in dev and realised I'd updated the email and address pages in the organisation folder, not the registration one. So this is those changes.


jakzal (approved):
Fixed a test for you 👍

andymantell (approved):

PR #859, opened 2024-10-28
added code to hide PPON and other based on conditions

shilpigoel-goaco:


Frontend/CO.CDP.OrganisationApp/Constants/OrganisationSchemeType.cs
```diff
@@ -1,26 +1,46 @@
...
-            { "GG-RCE", "Guernsey Registry Number"},
-            { "JE-FSC", "Jersey Financial Services Commission Registry Number"},
-            { "IM-CR", "Isle of Man Companies Registry Number"},
-            { "GB-NHS", "National health Service Organisations Registry Number"},
-            { "GB-UKPRN", "UK Register of Learning Provider Number"},
-            { "VAT", "VAT number"},
-            { "Other", "Other / None"},
-            { "GB-PPON", "Ppon" }
-        };
+    public const string CompaniesHouseNumber = "GB-COH";
```
jakzal, line 7: Naming is t consistent.  “Number” is only used in two cases. Can we drop it to keep scheme naming consistent?

jakzal (approved):
Looks good apart from a minor comment 👍

dharmverma (approved):

Frontend/CO.CDP.OrganisationApp/Constants/OrganisationSchemeType.cs
shilpigoel-goaco: Done!! renamed to CompanyHouse

jakzal: VATNumber suffers the same inconsistency

PR #864, opened 2024-10-28
Create a Common API Exception Handler Middleware

dharmverma:
Create a Common API Exception Handler Middleware


jakzal:
Looks good so far!

Libraries/CO.CDP.WebApi.Foundation/ResponseMiddleware.cs
```diff
@@ -0,0 +1,74 @@
...
+                logger.LogInformation("Response status: {statusCode}, for request: {requestUrl}", statusCode, requestUrl);
+
+                await HandleResponse(context, statusCode, new ProblemDetails { Status = statusCode });
+            }
+        }
+        catch (Exception ex)
+        {
+            var statusCode = StatusCodes.Status500InternalServerError;
+            var errorCode = "GENERIC_ERROR";
+            var pd = new ProblemDetails { Status = statusCode };
```
jakzal, line 35: Since we create a new instance of ProblemDetails here (we didn't prteviously), do we lose the default behaviour? Or maybe, problem details created here are merged with the defaults?

Libraries/CO.CDP.WebApi.Foundation/ResponseMiddleware.cs
```diff
@@ -0,0 +1,74 @@
...
+
+            if (exceptionMap.TryGetValue(ex.GetType(), out (int status, string code) error))
+            {
+                errorCode = error.code;
+                pd.Status = statusCode = error.status;
+                pd.Detail = ex.Message;
+                logger.LogInformation(ex, "Response status: {statusCode}, for request: {requestUrl}", statusCode, requestUrl);
```
jakzal, line 42: If the exception results in 50x error, we should log an error, information otherwise.

Libraries/CO.CDP.WebApi.Foundation/ResponseMiddleware.cs
```diff
@@ -0,0 +1,74 @@
...
+    {
+        var requestUrl = context.Request.GetDisplayUrl();
+
+        try
+        {
+            await next.Invoke(context);
+
+            var statusCode = context.Response.StatusCode;
+
+            if (statusCode >= 400 && statusCode < 500)
```
jakzal, line 24: When is this triggered? Is it when an error response is returned from a controller explicitly?

dharmverma: yes, like bad request, unprocessable entity, unauthorized, forbidden

Libraries/CO.CDP.WebApi.Foundation/ResponseMiddleware.cs
dharmverma: For uncaught exception (not in this middleware), we are logging them as error.
Do you think for known Error codes(our error map) where status is 500, should we log as exception?

Libraries/CO.CDP.WebApi.Foundation/ResponseMiddleware.cs
dharmverma: Yes that's the problem, and I am looking over internet for a solution.
And hence the draft PR :)

dharmverma: Oh I see it contains default behaviour, like setting title

PR #875, opened 2024-10-30
DP-741 - AddDbFunctionToUpdateQuestionsSortOrder added

maciej-goaco:


jakzal (approved):

maciej-goaco:
The order of questions in each section is defined by sort_order field.
It's calculated based on the next_question_id field value contaning the id of the next question.
That implies the following logic:
  - first question in each section would not have its ID referenced in any next_question_id field
  - last question in each section would have its next_question_id field set to NULL

To eliminate the need to manually recalculate the order of questions after inserting/deleting a question,
a SQL function update_form_questions_sort_order() has been introduce to automatically 
recalculate it for all sections and questions.

dharmverma (approved):

PR #880, opened 2024-10-31
DP-786 - Rejected/pending approval flow, and resending support admin approval emails

andymantell:


Frontend/CO.CDP.OrganisationApp/Pages/Organisation/OrganisationOverview.cshtml.cs
```diff
@@ -19,7 +19,7 @@ public async Task<IActionResult> OnGet()
     {
         try
         {
-            OrganisationDetails = await organisationClient.GetOrganisationAsync(Id);
+            OrganisationDetails = await organisationClient.GetOrganisationExtendedAsync(Id);
```
andymantell, line 22: Had to call a new "extended" endpoint here, since we want details about the review status of the org, not just the basic fields.

Frontend/CO.CDP.OrganisationApp/Pages/Organisation/OrganisationSelection.cshtml.cs
```diff
@@ -1,23 +1,35 @@
...
-        usersTenant.Tenants.ToList()
-            .ForEach(t => t.Organisations.ToList()
-                .ForEach(o => UserOrganisations.Add(o)));
+        // TODO: Tenant call doesn't return details of review comments - need to extend it to include a review_status column as a minimum as part of a refactor of reviews
+        // For the moment, hack this data in just for the intial proof of concept
+        var organisationTasks = usersTenant.Tenants
+            .SelectMany(t => t.Organisations)
+            .Select(async o => await organisationClient.GetOrganisationExtendedAsync(o.Id));
+
+        Organisations.AddRange(await Task.WhenAll(organisationTasks));
```
andymantell, line 27: Temporary hack to call the extended org endpoint to expose the review data since it doesn't come back from the tenant api.

If we had a simple `status` column returned from this we wouldn't have to do this.

Frontend/CO.CDP.OrganisationApp/WebApiClients/OrganisationExtensions.cs
```diff
@@ -20,6 +20,11 @@ public static bool IsPendingBuyer(this Organisation.WebApiClient.Organisation or
         return organisation.Details.PendingRoles.Contains(PartyRole.Buyer);
     }
 
+    public static bool IsRejectedBuyer(this Organisation.WebApiClient.Organisation organisation)
+    {
+        return IsPendingBuyer(organisation) && organisation.Details.Approval.ReviewedBy != null;
+    }
+
```
andymantell, line 27: Inferring rejected status by the fact that the buyer is pending _and_ has a non null review. Would be nice to have the review_status field explicitly recording this instead of having to infer it.

Services/CO.CDP.Organisation.WebApi/Api/Organisation.cs
```diff
@@ -101,6 +101,31 @@ await useCase.Execute(organisationId)
...
+                operation.Description = "Get an organisation by ID.";
+                operation.Summary = "Get an organisation by ID with additional details.";
+                operation.Responses["200"].Description = "Extended Organisation details.";
+                operation.Responses["401"].Description = "Valid authentication credentials are missing in the request.";
+                operation.Responses["404"].Description = "Organisation not found.";
+                operation.Responses["500"].Description = "Internal server error.";
+                return operation;
+            });
+
         app.MapPatch("/organisations/{organisationId}",
```
andymantell, line 129: The "extended" organisation endpoint - so I could fetch additional review data.

Maybe if we had this simple "review_status" column then we could remove this additional endpoint and just return it from the main one.

Services/CO.CDP.Organisation.WebApi/UseCase/UpdateOrganisationUseCase.cs
```diff
@@ -189,12 +196,71 @@ public async Task<bool> Execute((Guid organisationId, UpdateOrganisation updateO
...
+            (updateType == OrganisationUpdateType.OrganisationName || updateType == OrganisationUpdateType.OrganisationEmail)
+            && organisation.PendingRoles.Contains(PartyRole.Buyer)
+            && organisation.ReviewedBy != null
+        )
+        {
+            await ResendBuyerApprovalEmail(organisation);
+
+            // TODO: We're "resetting" the review status here but the only way is to wipe the previous review. We should likely refactor this to have a review_status column.
+            organisation.ReviewComment = null;
+            organisation.ReviewedBy = null;
```
andymantell, line 219: This is a bit grim. Again, if we had the nice "status" column then we would just reset it here. But instead I'm having to wipe out the previous review in order to reset their status from rejected back to pending approval

Frontend/CO.CDP.OrganisationApp/Pages/Organisation/OrganisationOverview.cshtml.cs
jakzal: discussed: let's call the new `GET /organisations/{organisationId}/approvals` here to enrich organisation details if it has pending roles

Frontend/CO.CDP.OrganisationApp/WebApiClients/OrganisationExtensions.cs
jakzal: This decision will be done on the API side where we will expose the status on the response.

JBaigGoaco (approved):

Services/CO.CDP.OrganisationInformation.Persistence/DatabaseOrganisationRepository.cs
```diff
@@ -13,6 +13,7 @@ public void Dispose()
     public async Task<Organisation?> Find(Guid organisationId)
```
dharmverma, line 13: This repo function is used at many places, do we need ReviewedBy all the time? 
Is it possible to create a separate function and use when required?

Services/CO.CDP.Organisation.WebApi/UseCase/UpdateOrganisationUseCase.cs
```diff
@@ -188,12 +195,70 @@ public async Task<bool> Execute((Guid organisationId, UpdateOrganisation updateO
...
+            await ResendBuyerApprovalEmail(organisation);
+
+            organisation.ReviewComment = null;
+            organisation.ReviewedBy = null;
+        }
+    }
+
+    private async Task ResendBuyerApprovalEmail(Persistence.Organisation organisation)
+    {
+        var baseAppUrl = configuration.GetValue<string>("OrganisationAppUrl") ?? "";
```
dharmverma, line 223: Would be nice, if we have all EmailNotification related logic (templateId, reading configs and building EmailNotificationRequest object) encapsulated away into a separate EmailService class, 
where all we need to call the method and required parameter like below - 

`EmailService.SendBuyerApprovalEmail(string organisationName, string requestLink);`

dharmverma: I understand this change may not lie under current story scope, but we can start moving this method and later other email notification methods one by one.

Services/CO.CDP.Organisation.WebApi/UseCase/GetReviewsUseCase.cs
```diff
@@ -0,0 +1,15 @@
...
+
+namespace CO.CDP.Organisation.WebApi.UseCase;
+public class GetReviewsUseCase(IOrganisationRepository organisationRepository, IMapper mapper) : IUseCase<Guid, IEnumerable<Model.Review>>
+{
+    public async Task<IEnumerable<Model.Review>> Execute(Guid organisationId)
+    {
+        var organisation = await organisationRepository.Find(organisationId)
+                           ?? throw new UnknownOrganisationException($"Unknown organisation {organisationId}.");
+
+        return [ mapper.Map<Review>(organisation) ];
```
dharmverma, line 13: If Review is not a collection probably just remove collection from method response type

andymantell: The idea here was to expose a collection of one item _now_, so that when the reviews API is extended later then we don't need to change the way we consume it from the frontend. We are anticipating that very soon we will need to have multiple reviews per org.

But, I do take your point. I don't know - what do you think @jakzal ?

Services/CO.CDP.Organisation.WebApi/UseCase/UpdateOrganisationUseCase.cs
andymantell: yeah the amount of duplication has got a bit out of hand there hasn't it. I'd be happy to do it, but yeah perhaps not in this change right now.

Next time we modify the email code, definitely...

dharmverma (approved):

dbgoaco (approved):

Frontend/CO.CDP.OrganisationApp.Tests/Pages/OrganisationSelectionTest.cs
```diff
@@ -27,34 +30,83 @@ public async Task OnGet_WhenSessionIsValid_ShouldPopulatePageModel()
...
 
-    private TenantLookup GetUserTenant()
+    private TenantLookup GetUserTenant(ICollection<PartyRole>? pendingRoles = null)
     {
         return new TenantLookup(
                 new List<UserTenant>
                 {
                     new(
-                        id: Guid.NewGuid(),
+                        id: new Guid(),
```
jakzal, line 101: Why this change?

andymantell: I suspect I changed something and then ended up putting it back. I will undo it

Services/CO.CDP.Organisation.WebApi.Tests/UseCase/GetReviewsUseCaseTest.cs
```diff
@@ -0,0 +1,96 @@
...
+using CO.CDP.Organisation.WebApi.UseCase;
+using CO.CDP.Organisation.WebApi.Model;
+using CO.CDP.OrganisationInformation.Persistence;
+
+namespace CO.CDP.Organisation.WebApi.Tests.UseCase;
+
+public class GetReviewsUseCaseTests
+{
+    private readonly Mock<IOrganisationRepository> _organisationRepositoryMock;
+    private readonly Mock<IMapper> _mapperMock;
```
jakzal, line 13: Our approach was not to mock automapper, but use it directly. This way this test becomes a test for the use case as well as mapping (automapper becomes implementation detail). Otherwise, we'd need to write separate tests for mapping. It's a compromise. 

Here's an example:

https://github.com/cabinetoffice/GCGS-Central-Digital-Platform/blob/ba7fb2ac9249cf434bf9f02c2362148cabfff766/Services/CO.CDP.Organisation.WebApi.Tests/UseCase/GetOrganisationUseCaseTest.cs#L11

Services/CO.CDP.Organisation.WebApi/Api/Organisation.cs
```diff
@@ -101,6 +101,31 @@ await useCase.Execute(organisationId)
...
 
+        app.MapGet("/organisations/{organisationId}/reviews",
+            [OrganisationAuthorize(
+                [AuthenticationChannel.OneLogin, AuthenticationChannel.ServiceKey],
+                [Constants.OrganisationPersonScope.Admin, Constants.OrganisationPersonScope.Editor, Constants.OrganisationPersonScope.Viewer],
+                OrganisationIdLocation.Path,
+                [Constants.PersonScope.SupportAdmin])]
+        async (Guid organisationId, IUseCase<Guid, IEnumerable<Review>> useCase) =>
+        await useCase.Execute(organisationId)
+                   .AndThen(organisation => organisation != null ? Results.Ok(organisation) : Results.NotFound()))
```
jakzal, line 112: Copy&Paste problem. This isn't an organisation we're dealing with here :D

Services/CO.CDP.Organisation.WebApi/UseCase/GetReviewsUseCase.cs
jakzal: Keep the collection, please. We've done this consciously, as we anticipate we will need multiple reviews per org. 

At the moment, reviewers can add reviews multiple times, but we always overwrite the previous review.

Services/CO.CDP.Organisation.WebApi.Tests/UseCase/GetReviewsUseCaseTest.cs
andymantell: Done

Services/CO.CDP.Organisation.WebApi/Api/Organisation.cs
andymantell: Done

Services/CO.CDP.OrganisationInformation.Persistence/DatabaseOrganisationRepository.cs
andymantell: @jakzal This is done, please can you confirm the approach in 65a493d71fcc3e073753aa98778093768c356e98 - as discussed

jakzal: Looks fine , though I think the same result might be achieved by replacing this:

```csharp
var review = mapper.Map<Review>(await organisationRepository.FindIncludingReviewedBy(organisation.Guid));
```

with this:

```csharp
var review = mapper.Map<Review>(organisation);
```

andymantell: I tried this, it doesn't work - the plain organisation object there doesn't have reviewed by on it. No magic happened...

PR #886, opened 2024-11-01
Dp 811 organisation receive multiple joining request

dbgoaco:
Prevents multiple org join requests for the same user. [DP-811](https://noticingsystem.atlassian.net/jira/software/projects/DP/boards/22?selectedIssue=DP-811)

### DIFF
--- a/Frontend/CO.CDP.OrganisationApp/Constants/ErrorMessagesList.cs
+++ b/Frontend/CO.CDP.OrganisationApp/Constants/ErrorMessagesList.cs
@@ -23,7 +23,7 @@ public static class ErrorMessagesList
     public const string UnknownBuyerInformationUpdateType = "The requested buyer information update type is unknown.";
 
     public const string DuplicateApiKeyName = "An API key with this name already exists. Please try again.";
-    public const string AlreadyMemberOfOrganisation = "You are already a member of this organisation.";
+    public const string AlreadyMemberOfOrganisation = "You are either already a member of this organisation or have already requested to join.";
 
     public const string OrganisationInvalidInput = "Invalid Input.";
     public const string MissingOrganisationName = "Missing organisation name.";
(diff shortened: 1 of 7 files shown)


Frontend/CO.CDP.OrganisationApp/Constants/ErrorMessagesList.cs
```diff
@@ -23,7 +23,7 @@ public static class ErrorMessagesList
     public const string UnknownBuyerInformationUpdateType = "The requested buyer information update type is unknown.";
 
     public const string DuplicateApiKeyName = "An API key with this name already exists. Please try again.";
-    public const string AlreadyMemberOfOrganisation = "You are already a member of this organisation.";
+    public const string AlreadyMemberOfOrganisation = "You are either already a member of this organisation or have already requested to join.";
```
jakzal, line 26: Did we run this wording through @jason-goaco ? :)

Services/CO.CDP.OrganisationInformation.Persistence/DatabaseOrganisationJoinRequestRepository.cs
```diff
@@ -33,6 +33,12 @@ public async Task<IEnumerable<OrganisationJoinRequest>> FindByOrganisation(Guid
             .ToArrayAsync();
     }
 
+    public async Task<OrganisationJoinRequest?> FindByOrganisationAndPerson(Guid organisationId, int personId)
```
jakzal, line 36: We use `Guid` for organisation ID, while `int` for person id. Any particular reason for this?

Frontend/CO.CDP.OrganisationApp/Constants/ErrorMessagesList.cs
dbgoaco: @jason-goaco tweak to the working, here's why. Previously it contained the link to join the organisation but there are 3 scenarios that can play out here: 
a) user is not a member and has NOT already submitted a join request;
b) user is not a member but HAS submitted a join request;
c) user has previously attempted to join but was rejected.

To cater for all three scenarios with different wording would require quite a bit of effort based on the current code setup relative to the value added. I thought it made more sense to keep the wording of the message more generic.

Services/CO.CDP.OrganisationInformation.Persistence/DatabaseOrganisationJoinRequestRepository.cs
dbgoaco: No reason, updated to Guid.

dpatel017 (approved):

PR #888, opened 2024-11-01
DP-755 - Fixed migration added

maciej-goaco:


andymantell (dismissed):

andymantell (changes requested):
```Npgsql.PostgresException : 42P13: cannot change return type of existing function```

Looks like replacing it won't work. Can you remove the function and re-add it?

andymantell (approved):

jakzal (approved):

PR #895, opened 2024-11-04
DP-734 - Connected persons PCS name not pulling through in H1

dpatel017:
Connected persons PCS name not pulling through in H1

### DIFF
--- a/Frontend/CO.CDP.OrganisationApp/Pages/Supplier/ConnectedEntity/ConnectedEntityCompanyRegisterName.cshtml.cs
+++ b/Frontend/CO.CDP.OrganisationApp/Pages/Supplier/ConnectedEntity/ConnectedEntityCompanyRegisterName.cshtml.cs
@@ -81,7 +81,10 @@ public class ConnectedEntityCompanyRegisterNameModel(ISession session) : PageMod
     private void InitModal(ConnectedEntityState state, bool reset = false)
     {
         Caption = state.GetCaption();
-        Heading = $"Select where {state.OrganisationName} is registered as person with significant control";
+        var name = state.ConnectedEntityType == Constants.ConnectedEntityType.Organisation
+                    ? state.OrganisationName : state.FirstName;
+
+        Heading = $"Select where {name} is registered as person with significant control";
         ConnectedEntityType = state.ConnectedEntityType;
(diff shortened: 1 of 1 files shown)


Frontend/CO.CDP.OrganisationApp/Pages/Supplier/ConnectedEntity/ConnectedEntityCompanyRegisterName.cshtml.cs
```diff
@@ -81,7 +81,10 @@ public IActionResult OnPost()
     private void InitModal(ConnectedEntityState state, bool reset = false)
     {
         Caption = state.GetCaption();
-        Heading = $"Select where {state.OrganisationName} is registered as person with significant control";
+        var name = state.ConnectedEntityType == Constants.ConnectedEntityType.Organisation
+                    ? state.OrganisationName : state.FirstName;
```
andymantell, line 85: Is first name enough? I think we'd want last name too. Consider the scenario where the org has 5 connected persons, all called "John"

andymantell (changes requested):

Frontend/CO.CDP.OrganisationApp/Pages/Supplier/ConnectedEntity/ConnectedEntityCompanyRegisterName.cshtml.cs
dpatel017: Valid point.  I have asked Luke to clarify as other places we using only first name.

dpatel017: Also before coming to this screen user enters first name and last name, so technically they know which user they are entering or updating information.

andymantell: @dpatel017 I'll mark this as resolved. If it's consistent with the other screens then fine. And if it's insufficient, I'm sure it will get raised and we can address it as a whole unit.

andymantell (approved):

PR #933, opened 2024-11-14
Localizing forms texts in data sharing api and pdf

andymantell:


Services/CO.CDP.DataSharing.WebApi/Model/FormQuestion.cs
```diff
@@ -11,7 +11,7 @@ public record FormQuestion
     public required string Title { get; init; }
 
     /// <example><![CDATA["<span>Central Government Only - UK</span><br /><p>For contracts which relate to projects/programmes (i) with a value of £10 million or more; or (ii) a value of less than £10 million where it is anticipated that the project will require in excess of 500 tonnes of steel; please describe the steel specific supply chain management systems, policies, standards and procedures you have in place to ensure robust supply chain management and compliance with relevant legislation.</p>Please provide details of previous similar projects where you have demonstrated a high level of competency and effectiveness in managing all supply chain members involved in steel supply or production to ensure a sustainable and resilient supply of steel."]]></example>
-    public required string Text { get; init; }
+    public string? Text { get; init; }
```
andymantell, line 14: Had to change this since it is mapped from Description which is nullable on the source model in the mapper config

jakzal: We shouldn't be changing the data sharing contract if we can avoid it. Can we set it to an empty string during mapping?

JBaigGoaco (approved):

PR #965, opened 2024-11-21
DP-892: Re-introduce cache to tenant lookup calls

jakzal:


Frontend/CO.CDP.OrganisationApp/Pages/Shared/_Layout.cshtml
```diff
@@ -2,6 +2,20 @@
...
+        try
+        {
+            return await userInfoService.GetUserInfo();
+        }
+        catch
+        {
+            return null;
+        }
+    }
+}
```
andymantell, line 18: Is this supposed to be here? And this line a bit further down:

```
var userInfo = User.Identity?.IsAuthenticated == true ? await GetUserInfo() : null;
```

Feels like the kind of thing that could do with being reusable rather than sitting in Layout.cshtml?

andymantell (changes requested):

Frontend/CO.CDP.OrganisationApp/Pages/Shared/_Layout.cshtml
jakzal: Yes, it didn't feel right when I wrote it. Last-quick-Friday-fix.

I think the real problem is we use something that throws an exception in the template. My mistake was removing this: https://github.com/cabinetoffice/GCGS-Central-Digital-Platform/pull/965/commits/2979a429d3101a55c263e65f9ee5a1a3bb2f29c2#diff-57befefa70a93ad3b3a0d46a4ffa3f9c79456991ce8406d70ae022a7d22b56e1

jakzal: Brought back `HasTenant` as `HasOrganisations`

andymantell (approved):

JBaigGoaco (approved):

PR #973, opened 2024-11-25
Feature/dp 800 translations share my information

rmohammed-goaco:
https://noticingsystem.atlassian.net/browse/DP-800

Add translations for share my information


Services/CO.CDP.Localization/StaticTextResource.resx
```diff
@@ -1452,6 +1455,65 @@
   <data name="Organisation_OrganisationIdentification_ODS_Code_ErrorMessage" xml:space="preserve">
     <value>Enter the code</value>
   </data>
+  <data name="Organisation_ShareYourInformation_Title" xml:space="preserve">
```
andymantell, line 1458: I don't think these keys should start with `Organisation_` since they are not in the organisation folder. `ShareYourInformation_Title` would be sufficient here

andymantell (changes requested):

Services/CO.CDP.Localization/StaticTextResource.resx
rmohammed-goaco: Done

andymantell (approved):

dharmverma (approved):

PR #984, opened 2024-11-26
DP-860 - Email address validation made GDS-compliant

maciej-goaco:


Libraries/CO.CDP.Mvc.Validation/EmailAddressValidator.cs
```diff
@@ -0,0 +1,92 @@
...
+public static class EmailAddressValidator
+{
+    private static readonly string ValidLocalChars = @"a-zA-Z0-9.!#$%&'*+/=?^_`{|}~\-";
+    private static readonly string EmailRegexPattern = $@"^[{ValidLocalChars}]+@([^.@][^@\s]+)$";
+    private static readonly Regex HostnamePartRegex = new(@"^(xn|[a-z0-9]+)(-?-[a-z0-9]+)*$", RegexOptions.IgnoreCase);
+    private static readonly Regex TldPartRegex = new(@"^([a-z]{2,63}|xn--([a-z0-9]+-)*[a-z0-9]+)$", RegexOptions.IgnoreCase);
+
+    public static bool IsValid(string emailAddress)
+    {
+        emailAddress = StripAndRemoveObscureWhitespace(emailAddress);
```
jakzal, line 17: If I understand this correctly, we strip some characters for validation. Does it mean we then persist the email with those characters included?

andymantell (approved):

Libraries/CO.CDP.Mvc.Validation/EmailAddressValidator.cs
maciej-goaco: Sanitising the value before validation has been taken out of the validation process and applied separately with a ModelBinder. This way it has become reusable and when modified to be applicable on the Page model level, we could easily apply it solution-wide to improve security (as fallback for eliminating XSS, SQL injections etc. if any sneaked in due to not using MS HTML extensions).

Frontend/CO.CDP.OrganisationApp.Tests/Extensions/StringExtensionsTests.cs
```diff
@@ -0,0 +1,18 @@
...
+using FluentAssertions;
+
+namespace CO.CDP.OrganisationApp.Tests.Extensions;
+
+public class StringExtensionsTests
+{
+    [Theory]
+    [InlineData("validated-text")]
+    [InlineData("  \tvalidated-text \x0c ")]
+    [InlineData("\rv\u200Cal\u200Did\u200Bate\u2060d-\u180Etex\uFEFFt\uFEFF")]
```
jakzal, line 11: Can we add a case for repeated characters to make sure it doesn't just remove one?

i.e. `[InlineData("ab\u200Ccd\u200Cef\u200C")]`

Frontend/CO.CDP.OrganisationApp/Models/SanitisedStringModelBinder.cs
```diff
@@ -0,0 +1,18 @@
+using CO.CDP.OrganisationApp.Extensions;
+using Microsoft.AspNetCore.Mvc.ModelBinding;
+
+namespace CO.CDP.OrganisationApp.Models;
+
+public class SanitisedStringModelBinder : IModelBinder
```
jakzal, line 6: 👌

jakzal (approved):

SDGoaco (approved):
Working as expected in the Local Environment

PR #987, opened 2024-11-27
DP-893 - Fix issue where connected entity individual category page loses track of connected entity type on post

andymantell:


Frontend/CO.CDP.OrganisationApp/Pages/Supplier/ConnectedEntity/ConnectedEntityIndividualCategory.cshtml.cs
```diff
@@ -40,7 +39,15 @@ public IActionResult OnGet()
...
     {
         var state = session.Get<ConnectedEntityState>(Session.ConnectedPersonKey);
-        if (state == null || !ModelState.IsValid)
+
+        if (state == null)
+        {
+            return RedirectToPage("ConnectedEntitySupplierCompanyQuestion", new { Id });
+        }
+
+        ConnectedEntityType = state.ConnectedEntityType;
```
andymantell, line 48: This is the crux of the fix - `ConnectedEntityType` was not being assigned a value on post, so when it re-rendered the page because of validation errors, you were being presented with the options of trustees instead of individuals and so it would get itself in a loop, redirecting you to "here" because it would not flow through `GetRedirectLinkPageName` correctly

Frontend/CO.CDP.OrganisationApp/Pages/Supplier/ConnectedEntity/ConnectedEntityIndividualCategory.cshtml.cs
```diff
@@ -40,7 +39,15 @@ public IActionResult OnGet()
...
+
+        if (!ModelState.IsValid)
         {
             return Page();
         }
```
andymantell, line 53: This is the crux of the fix - ConnectedEntityType was not being assigned a value on post, so when it re-rendered the page because of validation errors, you were being presented with the options of trustees instead of individuals and so it would get itself in a loop, redirecting you to "here" because it would not flow through GetRedirectLinkPageName correctly

dpatel017 (approved):

MayuriGoaco (approved):
Tested lcoally.Issue is fixed.

PR #1003, opened 2024-12-02
Update date field test with a slightly more optimistic future date

andymantell:


webit4me (approved):
thanks Andy. 
I was hoping it could be dynamic, and even validate a day in future. but given 2099 I wont be in this planet, it has my blessing ;)

dbgoaco (approved):

PR #971, opened 2024-11-22
DP-768: Store sessions in Redis

mareg:
To test it locally, add the following to your `compose.override.yaml` file:
```yaml
  organisation-app:
    container_name: ""
    deploy:
      replicas: 10
```
This will start with 10 replicas for the `organisation-app` and `gateway` will send traffic equally between them.

PR #1007, opened 2024-12-02
DP-925 Set org-app session's flag and data protection prefix

webit4me:
image


terragrunt/modules/ecs/templates/task-definitions/organisation-app.json.tftpl
```diff
@@ -37,6 +37,8 @@
         {"name": "OrganisationService", "value": "https://organisation.${public_domain}"},
         {"name": "Organisation__Authority", "value": "https://authority.${public_domain}"},
         {"name": "PersonService", "value": "https://person.${public_domain}"},
+        {"name": "SharedSessions", "value": "${shared_sessions_enabled}"},
```
jakzal, line 40: Shared sessions is a feature flag, so should be something like `Features__SharedSessions`. @mareg please confirm.

Related app change: https://github.com/cabinetoffice/GCGS-Central-Digital-Platform/pull/971/files#diff-efd1dcae149754760876b9e248c8b832b93447145deca248bf3e35ea08efe348

PR #1029, opened 2024-12-09
Update POST /organisations to use the logged in user rather than the one passed in the payload

dharmverma:
dp-937 Update POST /organisations to use the logged in user rather than the one passed in the payload


jakzal (approved):

JBaigGoaco (approved):

MayuriGoaco (approved):

SDGoaco (commented):
working as expected

PR #1031, opened 2024-12-09
Switch address rendering to commas temporarily to avoid html rendering issues, following up from XSS fixes

andymantell:


dharmverma (approved):

MayuriGoaco (approved):
Tested locally.Now the postal address is not displaying the <br/> tags.Working as expected.

dpatel017 (approved):

PR #1032, opened 2024-12-09
dp-989 introduce database connection pool setting

dharmverma:
dp-989 introduce database connection pool setting

### DIFF
--- a/Libraries/CO.CDP.Configuration/Helpers/ConnectionStringHelper.cs
+++ b/Libraries/CO.CDP.Configuration/Helpers/ConnectionStringHelper.cs
@@ -4,7 +4,7 @@ namespace CO.CDP.Configuration.Helpers;
 
 public static class ConnectionStringHelper
 {
-    private static readonly string[] requiredKeys = new[] { "Server", "Database", "Username", "Password" };
+    private static readonly string[] requiredKeys = ["Server", "Database", "Username", "Password"];
 
     public static string GetConnectionString(IConfiguration configuration, string name)
     {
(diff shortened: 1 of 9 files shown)


shilpigoel-goaco (approved):

andymantell (approved):

Libraries/CO.CDP.Configuration/Helpers/ConnectionStringHelper.cs
```diff
@@ -27,7 +27,9 @@ public static string GetConnectionString(IConfiguration configuration, string na
         var database = connectionSettings["Database"];
         var username = connectionSettings["Username"];
         var password = connectionSettings["Password"];
+        var maxPoolSize = connectionSettings["MaxPoolSize"];
+        var maxPoolSizeString = string.IsNullOrWhiteSpace(maxPoolSize) ? "" : $"MaxPoolSize={maxPoolSize};";
```
webit4me, line 31: Could you please define a fallback limit here? It's better to ensure we operate within a defined range, rather than relying on other components to handle the absence of this parameter. This way, we can avoid unexpected behaviour if the environment variable isn't set.

webit4me:
Have we considered setting different limits for different services based on their load? For example, services with lighter workloads could be allocated fewer connections, while those under heavier loads could be given higher bandwidth to the database. This approach might help optimise resource usage and prevent bottlenecks for critical services.

Libraries/CO.CDP.Configuration/Helpers/ConnectionStringHelper.cs
dharmverma: Yes they are defined as default in appsetting.json

PR #1036, opened 2024-12-09
DP-925 Scale back services to one task to park work on shared sessions

webit4me:


dharmverma (approved):

jakzal (approved):

terragrunt/components/terragrunt.hcl
```diff
@@ -268,7 +268,7 @@ locals {
   }
 
   # @TODO (ABN) Remove me
-  desired_count_development    = 8
+  desired_count_development    = 1
```
andymantell, line 271: Does the comment need to go as well?

andymantell (changes requested):

terragrunt/components/terragrunt.hcl
webit4me: No, the comment should remain for the entire section. 
Once we’re done testing the shared session (and now the database connections as well), we’ll revert back to having all pre-production accounts use the same configuration.

PR #1052, opened 2024-12-12
DP-1003: Adopt git merge strategy

mareg:


andymantell (approved):
Approving this as a recommendation, based on the idea that we trial it for the next month or so

PR #1062, opened 2024-12-13
DP-960 - added cache control header

shilpigoel-goaco:


dpatel017 (approved):

Frontend/CO.CDP.OrganisationApp/CacheControlMiddleware.cs
```diff
@@ -0,0 +1,25 @@
...
+    public CacheControlMiddleware(RequestDelegate next)
+    {
+        _next = next;
+    }
+
+    public async Task Invoke(HttpContext context)
+    {
+        context.Response.OnStarting(() =>
+        {
+            if (!context.Response.Headers.ContainsKey("Cache-Control"))
```
andymantell, line 16: Please can you add an extra bit of logic here so that it does not add this header to the static assets - CSS, JS etc. (Anything that comes from wwwroot). I ran this locally and it is adding the header to our site.css.

andymantell: Apart from that, looks good. Thanks 👍🏻

andymantell (changes requested):

andymantell (approved):

SDGoaco (approved):
Working as expected in Local Environment

PR #1088, opened 2024-12-18
DP-712: Use the DataSource rather than a connection string to configure the database

jakzal:
Cherry-picked from https://github.com/cabinetoffice/GCGS-Central-Digital-Platform/pull/1075

Also applied in other projects, not only in EntityVerification and Organisation.WebApi.

This is the way that's recommended by npgsql. It will enable us to configure postgres enums properly.


dharmverma (approved):

shilpigoel-goaco:
@jakzal shouldn't it be apply to all webapis?

jakzal:
@shilpigoel-goaco yes, I thought about it the second i postged it ;)

dbgoaco (approved):

PR #1104, opened 2025-01-06
DP-1045 organisation parties persistence

dharmverma:
DP-1045 organisation parties persistence

### DIFF
--- a/Services/CO.CDP.OrganisationInformation.Persistence/OrganisationParty.cs
+++ b/Services/CO.CDP.OrganisationInformation.Persistence/OrganisationParty.cs
@@ -1,0 +1,33 @@
+using CO.CDP.EntityFrameworkCore.Timestamps;
+using CO.CDP.OrganisationInformation.Persistence.Forms;
+using System.ComponentModel.DataAnnotations.Schema;
+
+namespace CO.CDP.OrganisationInformation.Persistence;
+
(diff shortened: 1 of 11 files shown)


Services/CO.CDP.OrganisationInformation.Persistence/OrganisationParty.cs
```diff
@@ -0,0 +1,33 @@
...
+    public Organisation? ChildOrganisation { get; set; }
+    public SharedConsent? SharedConsent { get; set; }
+
+    public DateTimeOffset CreatedOn { get; set; }
+    public DateTimeOffset UpdatedOn { get; set; }
+}
+
+public enum OrganisationRelationship
+{
+    Consortium
```
jakzal, line 32: I don't know the answer to this question, but is the relationship a "Consortium" or an "InformalConsortium". Does it matter?

dpatel017 (approved):

Services/CO.CDP.OrganisationInformation.Persistence/OrganisationParty.cs
dharmverma: We have a flag on Organisation table called Organisation Type, that covers what type of Consortium it is
I believe for above enum it can be just Consortium, as it doesn't matter here whether its formal or informal in this table

PR #1126, opened 2025-01-13
DP-1058 - Characters: 'B' and 'C' are not allowed as the last character of PPON IDs

dbgoaco:
[DP-1058](https://noticingsystem.atlassian.net/browse/DP-1058)


andymantell (approved):

dharmverma (approved):

SDGoaco (approved):
The Ticket is working as expected in the Local Environment

PR #1135, opened 2025-01-15
DP-712 Lock the outbox processor to ensure messages are only processed once

jakzal:
A new postgres NOTIFY might arrive before the previous one was processed. To reduce the likelihood of publishing the same message more than once, the OutboxProcessor now uses a lock. 

If OutboxProcessor is run twice in short span of time, the second run will wait for a short time, and then give up. It's fine to give up, as the current run will handle all messages anyway.

The functionality is still disabled. To enabled it, the `OutboxListener` feature flag needs to be set to `true` in the `Organisation.WebApi` and `EntityVerification` projects.

### DIFF
--- a/Libraries/CO.CDP.MQ.Tests/Outbox/OutboxProcessorTest.cs
+++ b/Libraries/CO.CDP.MQ.Tests/Outbox/OutboxProcessorTest.cs
@@ -61,6 +61,57 @@ public class OutboxProcessorTest
         _outbox.Messages.Should().Contain(m => m.Id == 2 && m.Published == true);
     }
 
+    [Fact]
+    public async Task ItOnlyProcessesOneRequestAtATime()
+    {
+        var messages = new List<OutboxMessage>
+        {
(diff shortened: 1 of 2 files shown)


Libraries/CO.CDP.MQ.Tests/Outbox/OutboxProcessorTest.cs
```diff
@@ -8,7 +8,7 @@ namespace CO.CDP.MQ.Tests.Outbox;
 public class OutboxProcessorTest
 {
     private readonly Mock<IPublisher> _publisher = new();
-    private readonly Mock<IOutboxMessageRepository> _outbox = new();
+    private readonly InMemoryOutboxMessageRepository _outbox = new();
```
jakzal, line 11: I resorted to a hand-crafted test double here, as we're reaching the limitations of Moq. Even if coordination of all the required calls was possible on the outbox repository, it would make tests hard to follow and hard to change.

andymantell (approved):

maciej-goaco (approved):

PR #1164, opened 2025-01-20
Feature/dp 821 buyer approval tabs pagination

rmohammed-goaco:


Frontend/CO.CDP.OrganisationApp/Pages/Shared/PaginationPartialModel.cs
```diff
@@ -0,0 +1,27 @@
...
+        public int TotalItems { get; set; }
+
+        [Required]
+        public int PageSize { get; set; }
+
+        [Required]
+        public string OrganisationType { get; set; } = "buyer";
+
+        [Required]
+        public string Url { get; set; }
```
andymantell, line 20: Some Github code warnings here, and throughout the PR

Frontend/CO.CDP.OrganisationApp/Pages/Shared/PaginationPartialModel.cs
```diff
@@ -0,0 +1,27 @@
...
+        public int CurrentPage { get; set; }
+
+        [Required]
+        public int TotalItems { get; set; }
+
+        [Required]
+        public int PageSize { get; set; }
+
+        [Required]
+        public string OrganisationType { get; set; } = "buyer";
```
andymantell, line 17: Does OrganisationType need to be on this model? The rest of it looks generic. And I can't see that it's used in the template?

Frontend/CO.CDP.OrganisationApp/Pages/Support/Organisations.cshtml
```diff
@@ -1,108 +1,124 @@
...
+        <a class="govuk-tabs__tab" href="/support/organisations/supplier">
+            Supplier organisations
+        </a>
+    </li>
+</ul>
+
+<br/>
+
+<h2 class="govuk-heading-m">There are @(Model.TotalOrganisations)
+    @(Model.Type)@(Model.TotalOrganisations == 0 || Model.TotalOrganisations > 1 ? "s" : "")</h2>
```
andymantell, line 31: There's a few things need localizing here

Frontend/CO.CDP.OrganisationApp/README.md
```diff
@@ -0,0 +1,53 @@
+# CO.CO.CDP.OrganisationInformation.OrganisationApp
```
andymantell, line 1: Great to see some documentation for this... :-)
But I wonder if it's maybe in the wrong place? Could it maybe sit inside `docs/development/cookbooks` ?

andymantell (changes requested):

Frontend/CO.CDP.OrganisationApp/README.md
rmohammed-goaco: Sure. Seems like its a place that no-one will ever find, but I'll change it

Frontend/CO.CDP.OrganisationApp/Pages/Support/Organisations.cshtml
rmohammed-goaco: I have no idea how to go about localising the above line. Could you help with this?

rmohammed-goaco: Ignore my last message. Sorted it

Frontend/CO.CDP.OrganisationApp/Pages/Support/Organisations.cshtml
```diff
@@ -1,108 +1,127 @@
...
+            @StaticTextResource.Support_Organisations_SupplierOrganisations_Title
+        </a>
+    </li>
+</ul>
+
+<br/>
+
+<h2 class="govuk-heading-m">
+    @StaticTextResource.Support_OrganisationApproval_there_are @(Model.TotalOrganisations)
+    @(Model.Type == "buyer" ? @StaticTextResource.Support_OrganisationApproval_buyers : @StaticTextResource.Support_OrganisationApproval_suppliers)
```
andymantell, line 32: Because of the potential structural differences between English and Welsh, this should be a 2 translatable strings, with data injected into it, like:

There are {0} buyers
There are {0} suppliers

At the moment the above heading is the result of concatenating the words "There are", the number, and the word "buyer" or "supplier" but that doesn't give the translation team much flexibility.

For example, the Welsh for "buyers" is "prynwyr" but "there are 3 buyers" is "mae 3 phrynwr". Note how "prynwyr" is spelt differently to "phrynwr". Therefore translating words in isolation without context and concatenating them is not sufficient. Remember that the translation team will just get a spreadsheet which asks them to translate the word "buyer" in isolation - they won't see the context in which it is going to be used. So it's important that the full translatable string contains all the context - with data represented as placeholders

(For ease, don't bother handling the case where there's "1 buyer" - because we shouldn't be in that situation for very long 😆)

Frontend/CO.CDP.OrganisationApp/Pages/Support/Organisations.cshtml
```diff
@@ -1,108 +1,127 @@
...
-        <a class="govuk-link" href="/support/organisations/buyer">@StaticTextResource.Support_Organisations_BuyerOrganisations_Link</a>
-    }
-}
+
+@await Html.PartialAsync("_Pagination", new PaginationPartialModel
+{
+    CurrentPage = Model.CurrentPage,
+    TotalItems = Model.TotalOrganisations,
+    PageSize = Model.PageSize,
+    Url = Model.PageUrl
```
andymantell, line 126: Still some Github code warnings down here.

Services/CO.CDP.Organisation.WebApi/Model/Command.cs
```diff
@@ -334,9 +334,14 @@ public record SupportOrganisationInfo
...
-    public int? Limit { get; init; }
-    public int? Skip { get; init; }
+    public string Type { get; init; }
+    public int Limit { get; init; }
+    public int Skip { get; init; }
+}
+
+public record OrganisationTypeQuery
+{
+    public string Type { get; init; }
```
andymantell, line 344: More github code warnings down here

Services/CO.CDP.OrganisationInformation.Persistence/DatabaseOrganisationRepository.cs
```diff
@@ -160,6 +160,50 @@ public async Task<IList<Organisation>> Get(string? type)
...
+    public async Task<int> GetTotalCount(string? type)
+    {
+        IQueryable<Organisation> result = context.Organisations;
+
+        switch (type)
+        {
+            case "buyer":
+                result = result.Where(o => o.Roles.Contains(PartyRole.Buyer) || o.PendingRoles.Contains(PartyRole.Buyer));
+                break;
+            case "supplier":
```
andymantell, line 199: You're taking "supplier" as the argument and mapping it to tenderer internally, which I understand. But I feel like the API should stick to referring to them as tenderers?

What I'm trying to say is - I know we call them suppliers and present them as such in the UI - but internally we should be consistent about referring to them as tenderers. Especially as supplier is another, equally valid PartyRole. Mapping supplier to tenderer in this api endpoint confuses the issue.

andymantell (changes requested):

Services/CO.CDP.OrganisationInformation.Persistence/DatabaseOrganisationRepository.cs
rmohammed-goaco: I'm going to push back on this one because I'm using the value which is passed from url and is front facing.

andymantell: I can see what you mean. I think the API should remain true to our internal representation though so that we at least isolate the confusion to the frontend app. So you could do the mapping before you call the API, instead of _inside_ the API?

andymantell: Discussed in a DM - have found an existing instance of this supplier -> tenderer mapping in the `Get` method in this same `DatabaseOrganisationRepository.cs` file.

I don't super love it, but what you're doing here is consistent with what we have already, so for now I think we can roll with it and if it becomes an issue in future we can refactor it out.

rmohammed-goaco:
@andymantell Tests will fail. Before I update them please let me know if you are happy with the new approach

Services/CO.CDP.OrganisationInformation.Persistence/DatabaseOrganisationRepository.cs
```diff
@@ -160,6 +160,50 @@ public async Task<IList<Organisation>> Get(string? type)
...
+        IQueryable<Organisation> result = context.Organisations;
+
+        switch (role)
+        {
+            case PartyRole.Buyer:
+                result = result.Where(o => o.Roles.Contains(PartyRole.Buyer) || o.PendingRoles.Contains(PartyRole.Buyer));
+                break;
+            default:
+                result = result.Where(o => o.Roles.Contains(PartyRole.Tenderer));
+                break;
```
andymantell, line 201: I think these API endpoints should take an array of `roles[]`, and the condition in the query would be like:

```
.Where(o => o.Roles.Any(role => roles.Contains(role)))
```

and just remove the switch entirely.

Currently this means that if you query for a list of orgs with the Buyer role, you get pending buyers back as well.

andymantell (changes requested):

Services/CO.CDP.Organisation.WebApi.Tests/UseCase/GetOrganisationsUseCaseTest.cs
```diff
@@ -62,30 +62,30 @@ public async Task Execute_WithValidCommand_ReturnsMappedOrganisations()
...
 
-        _organisationRepositoryMock.Setup(repo => repo.Get(command.Type)).ReturnsAsync(organisations);
+        _organisationRepositoryMock.Setup(repo => repo.GetPaginated(command.Role, command.Role, command.Limit, command.Skip)).ReturnsAsync(organisations);
         _mapperMock.Setup(m => m.Map<IEnumerable<OrganisationExtended>>(organisations)).Returns(mappedOrganisations);
 
         var result = await _useCase.Execute(command);
 
         result.Should().BeEmpty();
-        _organisationRepositoryMock.Verify(repo => repo.Get(command.Type), Times.Once);
+        _organisationRepositoryMock.Verify(repo => repo.GetPaginated(command.Role, command.Role, command.Limit, command.Skip), Times.Once);
```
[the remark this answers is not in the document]
rmohammed-goaco: Done

Services/CO.CDP.Localization/StaticTextResource.resx
```diff
@@ -1940,6 +1952,15 @@
...
   </data>
+  <data name="Support_OrganisationApproval_there_are" xml:space="preserve">
+    <value>There are</value>
+  </data>
+  <data name="Support_OrganisationApproval_buyers" xml:space="preserve">
+    <value>buyers</value>
+  </data>
+  <data name="Support_OrganisationApproval_suppliers" xml:space="preserve">
+    <value>suppliers</value>
+  </data>
```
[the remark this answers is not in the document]
rmohammed-goaco: Removed

andymantell (approved):

SDGoaco (approved):
Tested in Local Environment

PR #1173, opened 2025-01-22
DP-950: Remove obsolete code

mareg:


dbgoaco (approved):

maciej-goaco (approved):

andymantell (changes requested):
Can we keep this in for the moment? There's still login issues occasionally on integration and this might yet be useful debug info.

We can remove this eventually, but I think not yet... Have spoken to Dharm as well and he would like to keep it in for now

andymantell (approved):

PR #1206, opened 2025-01-28
DP-1093 added Buyer Organisation Type Hint text

shilpigoel-goaco:


Services/CO.CDP.Localization/StaticTextResource.resx
```diff
@@ -3226,6 +3214,18 @@
   <data name="MoU_SignedOn" xml:space="preserve">
     <value>Agreed on {0}</value>
   </data>
+  <data name="Footer_AboutUs" xml:space="preserve">
+    <value />
+  </data>
```
andymantell, line 3219: Can we remove this empty one? I think this happened on another PR too - I think the welsh version of it was still hanging around. If you remove the welsh version as well as this one, it should stop happening...

andymantell (changes requested):

Services/CO.CDP.Localization/StaticTextResource.resx
shilpigoel-goaco: done

shilpigoel-goaco: removed from both file.

andymantell (approved):

dpatel017 (approved):

MayuriGoaco (approved):

PR #1234, opened 2025-02-03
DP-1143 Data Sharing API: organisation for Individual supplier data Null

shilpigoel-goaco:


andymantell (approved):

MayuriGoaco (approved):
Tested locally.Now the datasharing API response returning null values for scale and other properties if Supplier organisation is a type of Individual.Working as Expected.

PR #1267, opened 2025-02-11
DP-1158-Buyers - Change Buyer Type

shilpigoel-goaco:


andymantell (changes requested):

Frontend/CO.CDP.OrganisationApp/Pages/Organisation/BuyerOrganisationType.cshtml.cs
```diff
@@ -0,0 +1,88 @@
...
+
+    public async Task<IActionResult> OnPost()
+    {
+        if (!ModelState.IsValid)
+        {
+            return Page();
+        }
+
+        var buyerInformation = await organisationClient.GetOrganisationBuyerInformationAsync(Id);
+        if (buyerInformation == null) return Redirect("/page-not-found");
```
andymantell, line 61: Wouldn't this one throw an exception instead of returning null? You could maybe just add this to the try catch below?

shilpigoel-goaco: No, it returns Null. 
image

andymantell: The use case returns null, but when that is invoked by the api endpoint, it throws a 404 when the result is null

{24C10529-831A-4E67-82DC-9660513B6A11}

andymantell (approved):

dpatel017 (approved):

MayuriGoaco (approved):

PR #1280, opened 2025-02-12
DP-1119 - Handle session timeouts more gracefully for users

dpatel017:
DP-1119 - Handle session timeouts more gracefully for users


Frontend/CO.CDP.OrganisationApp/Program.cs
```diff
@@ -279,6 +280,7 @@
 builder.Services.AddAuthorization();
 
 builder.Services.AddScoped<CookieAcceptanceMiddleware>();
+//builder.Services.AddScoped<DisplayLogoutMessageMiddleware>();
```
andymantell, line 283: I assume this isn't necessary?

andymantell (changes requested):

Frontend/CO.CDP.OrganisationApp/Program.cs
dpatel017: Yes, I have removed it.

dharmverma (approved):

PR #1291, opened 2025-02-14
Feature/dp 1178 organisation type scale and otherproperties updated

shilpigoel-goaco:


andymantell (changes requested):
This seems to have all the commits from #1267 in it as well? Could you create a new branch with only the Dp-1178 work on it?

I think this looks like the only bit of work that is relevant to DP-1178:

https://github.com/cabinetoffice/GCGS-Central-Digital-Platform/pull/1291/commits/88bd192f7b8ca5164f1b2c4ab6c801dd7c13b23b

So if you create a new branch off main, and then do `git cherry-pick 88bd192f7b8ca5164f1b2c4ab6c801dd7c13b23b` you should end up with a branch with just this work on it

PR #1294, opened 2025-02-14
DP-1178-Organisation Type: Scale and other properties are not getting…

shilpigoel-goaco:
… updated when Supplier is converted from type organisation to individual


dharmverma (approved):

MayuriGoaco (approved):

PR #1297, opened 2025-02-14
DP-1147, DP-1148 Expose connected person details in ShareCode endpoint

dharmverma:
DP-1147, DP-1148 Expose connected person details in ShareCode endpoint

### DIFF
--- a/Services/CO.CDP.DataSharing.WebApi/Model/AssociatedPerson.cs
+++ b/Services/CO.CDP.DataSharing.WebApi/Model/AssociatedPerson.cs
@@ -15,4 +15,27 @@ public record AssociatedPerson
     /// <example>"https://cdp.cabinetoffice.gov.uk/persons/c16f9f7b-3f10-42db-86f8-93607b034a4c"</example>
     public Uri? Uri { get; init; }
     [Required] public required List<PartyRole> Roles { get; init; }
+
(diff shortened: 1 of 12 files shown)


Services/CO.CDP.DataSharing.WebApi/Model/AssociatedPerson.cs
```diff
@@ -15,4 +15,27 @@ public record AssociatedPerson
...
+    public string? Nationality { get; set; }
+    public string? ResidentCountry { get; set; }
+    public IEnumerable<ControlCondition> ControlCondition { get; set; } = [];
+    public ConnectedPersonType ConnectedType { get; set; }
+
+
+    public IEnumerable<Address> Addresses { get; set; } = [];
+    public DateTimeOffset? RegisteredDate { get; set; }
+    public string? RegistrationAuthority { get; set; }
+    public bool HasCompnayHouseNumber { get; set; }
```
andymantell, line 36: If this is now being exposed via the data sharing api, I think it might be time to fix the typo that we've had for a while. If we don't do it now we'll never be able to do it!

`HasCompnayHouseNumber` => `HasCompanyHouseNumber`

dharmverma: Nice one, fixed now

andymantell (approved):

PR #1307, opened 2025-02-17
Support Admin - Search Pagination for Support Screen

rmohammed-goaco:
Nice simple solution to ensuring the search value is persisted even when navigating through the pages. Saves the search value to the session until it is cleared by the user.

### DIFF
--- a/Frontend/CO.CDP.OrganisationApp/Pages/Support/Organisations.cshtml.cs
+++ b/Frontend/CO.CDP.OrganisationApp/Pages/Support/Organisations.cshtml.cs
@@ -34,6 +34,8 @@ public class OrganisationsModel(
     {
         InitModel(type, pageNumber);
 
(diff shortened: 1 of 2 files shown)


Frontend/CO.CDP.OrganisationApp/Pages/Support/Organisations.cshtml.cs
```diff
@@ -43,6 +48,15 @@ public async Task<IActionResult> OnPost(string type, int pageNumber = 1)
     {
         InitModel(type, pageNumber);
 
+        if (!string.IsNullOrWhiteSpace(OrganisationSearchInput))
+        {
+            _session.Set("OrganisationSearchInput", OrganisationSearchInput);
```
dharmverma, line 53: You can use SessionContext from base class rather creating another variable of it

dharmverma (approved):

MayuriGoaco (approved):

PR #1311, opened 2025-02-18
DP-1209 - content for buyer org type page - main hint text and other radio label and hint text update

jason-goaco:


dharmverma (approved):

Frontend/CO.CDP.OrganisationApp/Pages/Registration/BuyerOrganisationType.cshtml
```diff
@@ -102,6 +102,9 @@
                             <label class="govuk-label govuk-radios__label" for="ca-type-5">
                                 @StaticTextResource.OrganisationRegistration_BuyerOrganisationType_Other_Label
                             </label>
+                             <div id="ca-type-5-item-hint" class="govuk-hint govuk-radios__hint">
```
andymantell, line 105: We shouldn't need Html.Raw here. Also this change needs applying to the _edit_ screen for this stuff.

jason-goaco: Ah soz. I should've asked but I thought copying the code above would do the trick.

jason-goaco: OK. I think I've sorted this based on our chat @andymantell. 
-  removed Html.Raw and brackets from the reg/BuyerOrganisationType.cshtml file
- made the change (excluding Html.Raw and brackets in the org/BuyerOrganisationType.cshtml file

PR #1332, opened 2025-02-20
DP-1177 - Requests to Join - Supplier Org Name Check

dpatel017:
Requests to Join - Supplier Org Name Check


Frontend/CO.CDP.OrganisationApp/Pages/Registration/OrganisationDetailsSummary.cshtml.cs
```diff
@@ -66,12 +68,39 @@ public async Task<IActionResult> OnPost()
         }
         catch (OrganisationWebApiClient.ApiException<OrganisationWebApiClient.ProblemDetails> aex)
         {
-            ApiExceptionMapper.MapApiExceptions(aex, ModelState);
+            var errorcode = ExtractErrorCode(aex);
+
+            if (errorcode == ErrorCodes.ORGANISATION_ALREADY_EXISTS)
+            {
+                var organisations = await organisationClient.SearchOrganisationAsync(RegistrationDetails.OrganisationName, null, 10,0.3);
```
andymantell, line 75: I think we should probably be using `LookupOrganisationAsync` not search, which invokes the trigram similarity and returns 10 results in this instance.

And in the `organisationRepository.FindByName` method, just make sure it's lowercasing the values before comparison - so that it ends up being a case insensitive match.

andymantell (changes requested):

MayuriGoaco (approved):

andymantell (approved):

rmohammed-goaco (approved):

MayuriGoaco (approved):

PR #1351, opened 2025-02-24
DP-1009 - MOU - Buyer Registration - Notify Email-  MoU

dpatel017:
DP-1009 - MOU - Buyer Registration - Notify Email-  MoU


rmohammed-goaco (approved):

Frontend/CO.CDP.OrganisationApp/CO.CDP.OrganisationApp.csproj
```diff
@@ -34,12 +34,6 @@
         <ProjectReference Include="..\..\Services\CO.CDP.Localization\CO.CDP.Localization.csproj" />
     </ItemGroup>
 
-    <ItemGroup>
-        <None Update="mou-pdfs\version-1.pdf">
-            <CopyToOutputDirectory>Always</CopyToOutputDirectory>
-        </None>
-    </ItemGroup>
-
```
andymantell, line 42: I see you've moved the MoU into www-root... Which would be fine in principle, but I think this will error in the `OnGetDownload` method of `Frontend\CO.CDP.OrganisationApp\Pages\MoU\ReviewAndSignMemorandom.cshtml.cs`

This loads the file up from disk and streams it out.

Also, if you look at this method that streams it out, it gives it a name dynamically on the fly, so whatever version of the MoU it currently is, it'll still be called `fts-joint-controller-agreement.pdf`

I think there's a couple of options:

1) Rework the email so that it points to this `OnGetDownload` action and just reuse that.

2) Consider - should the email download the version of the MoU that they _signed_? Probably yes? So we'd need a version of `OnGetDownload` that let you grab the specific version?

MayuriGoaco (approved):

Services/CO.CDP.Organisation.WebApi/UseCase/CreateOrganisationSignMouUseCase.cs
```diff
@@ -38,6 +41,63 @@ public async Task<bool> Execute((Guid organisationId, SignMouRequest signMouRequ
...
+        if (string.IsNullOrEmpty(baseAppUrl)) missingConfigs.Add("OrganisationAppUrl");
+        if (string.IsNullOrEmpty(templateId)) missingConfigs.Add("GOVUKNotify:MouDataSharingAgreementEmailTemplateId");
+
+        if (missingConfigs.Count != 0)
+        {
+            logger.LogError(new Exception("Unable to send MoU email"), $"Missing configuration keys: {string.Join(", ", missingConfigs)}. Unable to send MoU email to user.");
+            return;
+        }
+
+        var requestLink = new Uri(new Uri(baseAppUrl), $"/organisation/{organisation.Guid}/review-and-sign-memorandom?handler=download").ToString();
```
andymantell, line 65: This link will always fetch the latest copy of the MoU I believe. But if they accessed this email in 1 month's time, and we have published a new MoU in the meantime, we would need this link to download the copy of the MoU at the time that they signed it.

andymantell (changes requested):

andymantell (approved):

PR #1358, opened 2025-02-25
DP-1250 Allow same vat for multiple orgs

dharmverma:
DP-1250 Allow same vat for multiple orgs


dbgoaco (approved):

andymantell (approved):

MayuriGoaco (approved):
Tested locally.Now the same VAT can be used by multiple orgs.

PR #1382, opened 2025-02-28
DP-1264 - Handle users who delete their one login account create a new one with the same email (Resulting in new URN)

andymantell:


dharmverma (approved):

rmohammed-goaco (approved):
Looks like a sensible solution to me

MayuriGoaco (approved):
Tested locally.This issue is fixed.

PR #1389, opened 2025-03-04
DP-971 - Slimming down the containers to use the dotnet chiseled base image

andymantell:


compose.yml
```diff
@@ -127,7 +127,7 @@ services:
       redis:
         condition: service_healthy
     healthcheck:
-      test: ["CMD-SHELL", "curl --fail --silent --write-out '{\"status\":%{http_code}}' --output /dev/null http://127.0.0.1:8080/health"]
+      test: ["CMD", "nc", "-z", "127.0.0.1", "8080"]
```
webit4me, line 130: Make sure the health check:
 - Hits the `/health` endpoint
 - Get a 'HTTP/1.1 200' in response (not just a successful connection)
 
I have tested this locally against data-sharing service, but needs a proper test once applied on all services 

```suggestion
      test: ["CMD", "sh", "-c", "echo -e 'GET /health HTTP/1.1\\r\\nHost: localhost\\r\\nConnection: close\\r\\n\\r\\n' | nc -w 5 -q 1 127.0.0.1 8080 | grep 'HTTP/1.1 200'"]
```

dharmverma (approved):

rmohammed-goaco (approved):

PR #1406, opened 2025-03-10
Bugfix for: DP-1189 - Make user role descriptions specific to each user group (manage users)

rmohammed-goaco:
The content was displaying incorrectly for a pending buyer. This PR fixes that by adding the pending roles to do the content check against.

### DIFF
--- a/Services/CO.CDP.Organisation.WebApi/Model/Organisation.cs
+++ b/Services/CO.CDP.Organisation.WebApi/Model/Organisation.cs
@@ -27,7 +27,9 @@ public record Organisation
     /// <example>["supplier"]</example>
     public required List<PartyRole> Roles { get; init; }
 
-    public required Details Details { get; init; }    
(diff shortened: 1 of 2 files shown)


Services/CO.CDP.Organisation.WebApi/Model/Organisation.cs
```diff
@@ -27,7 +27,9 @@ public record Organisation
     /// <example>["supplier"]</example>
     public required List<PartyRole> Roles { get; init; }
 
-    public required Details Details { get; init; }    
+    public required List<PartyRole> PendingRoles { get; init; }
+
+    public required Details Details { get; init; }
```
andymantell, line 32: `PendingRoles` is already in `Details` - why do we need to surface it here?

andymantell (changes requested):

andymantell (approved):

dpatel017 (approved):

PR #1422, opened 2025-03-12
DP-1238 - Buyer - Change devolved Regs

dpatel017:
DP-1238 - Buyer - Change devolved Regs


andymantell (approved):

MayuriGoaco (approved):
Tested locally.Working as expected

PR #1444, opened 2025-03-19
Dp 1195 - Fix buyer search being bypassed when lookup endpoint throws a 404, but search endpoint returns results

andymantell:


MayuriGoaco (approved):

dharmverma (approved):

PR #1453, opened 2025-03-24
BAU Scale up migration services in all accounts

webit4me:
- Double of the size of each account's service instances


andymantell (approved):

andymantell:
Closing this one - we agreed it was not necessary and was born out of some confusion around migration services vs DB CPU usage.

PR #1468, opened 2025-03-27
Support Admin - Manage Users

rmohammed-goaco:


andymantell:
Useful to refer back to, but this PR is a dead end since we decided instead to allow support admins to do everything on orgs - we don't need to add this extra layer of granularity at the moment

PR #1474, opened 2025-03-28
Fixed organisation has registered companies house number not being set from true to false

rmohammed-goaco:


Frontend/CO.CDP.OrganisationApp/Pages/Supplier/ConnectedEntity/ConnectedEntityCheckAnswersOrganisation.cshtml.cs
```diff
@@ -210,6 +210,11 @@ private void InitModel(ConnectedEntityState state)
             addresses.Add(AddAddress(state.PostalAddress, CO.CDP.Organisation.WebApiClient.AddressType.Postal));
         }
 
+        if (state.SupplierHasCompanyHouseNumber!.Value == false)
+        {
+            state.CompaniesHouseNumber = null;
+        }
```
andymantell, line 216: `SupplierHasCompanyHouseNumber` is nullable - `bool?`

But if it is _actually_ null in reality, then `if (state.SupplierHasCompanyHouseNumber!.Value == false)` would throw a null reference exception.

Simpler to just check like this then we only nullify the  `CompaniesHouseNumber` when actually it's false? If it's null, it simply won't fall into this if statement because `null != false`

```suggestion
        if (state.SupplierHasCompanyHouseNumber == false)
        {
            state.CompaniesHouseNumber = null;
        }
```

andymantell (changes requested):

andymantell (approved):

Frontend/CO.CDP.OrganisationApp/Pages/Supplier/ConnectedEntity/ConnectedEntityCompanyQuestion.cshtml.cs
```diff
@@ -68,7 +68,7 @@ public IActionResult OnPost()
         }
 
         state.HasCompaniesHouseNumber = HasCompaniesHouseNumber;
-        state.CompaniesHouseNumber = CompaniesHouseNumber;
+        state.CompaniesHouseNumber = (HasCompaniesHouseNumber == true ? CompaniesHouseNumber : string.Empty);
```
rmohammed-goaco, line 71: Annoyingly I can't approve the PR, but I think you can @dpatel017 . Can you APPROVE on my behalf?

andymantell (approved):

MayuriGoaco (approved):

PR #1480, opened 2025-04-02
DP-1210 MoU email reminder job

dharmverma:
- Created a standalone "CO.CDP.ScheduledWorker" project
- Running MoU email reminder hourly
- Overlapping job handling


andymantell (approved):

Services/CO.CDP.ScheduledWorker/CompleteMoUReminderService.cs
```diff
@@ -0,0 +1,72 @@
...
+    {
+        logger.LogInformation("Complete MoU Reminder Processing Service started.");
+
+        var reminderList = await mouRepository.GetMouReminderOrganisations(DaysBetweenEmailReminders);
+
+        foreach (var organisation in reminderList)
+        {
+            await NotifyCompleteMoUReminder(organisation);
+
+            await mouRepository.UpsertMouEmailReminder(organisation.Id);
```
andymantell, line 24: @dharmverma Thinking about this - I think it should only update the MoU email reminder table if sending the emails was actually successful? Then if they fail, it would retry them on the next run? What do you think?

dharmverma: Good point @andymantell, updated the code to accommodate requirement.

dbgoaco (approved):

PR #1508, opened 2025-04-10
DP-1374 - Fix Support admin screen is showing duplicate organisations

rmohammed-goaco:
https://noticingsystem.atlassian.net/browse/DP-1374


andymantell (approved):
Tested locally and is working - orgs no longer showing as duplicate per admin email, and the list of admin emails is displaying nicely as a list

PR #1539, opened 2025-04-28
Bug fix: DP-1402 - Support admin search

dbgoaco:
[DP-1402](https://noticingsystem.atlassian.net/jira/software/projects/DP/boards/22?selectedIssue=DP-1402)


Services/CO.CDP.OrganisationInformation.Persistence/DatabaseOrganisationRepository.cs
```diff
@@ -271,8 +271,7 @@ public async Task<int> GetTotalCount(PartyRole? role, PartyRole? pendingRole, st
 
         if (searchText != null)
         {
-            result = result.Where(o => o.Name.Contains(searchText) ||
-                                       EF.Functions.TrigramsSimilarity(o.Name, searchText) > 0.3);
+            result = result.Where(o => o.Name.ToLower().Contains(searchText.ToLower()));
```
andymantell, line 274: I think this logic still doesn't sufficiently match what's in `GetPaginated` ? So the count will not match the number of results.

andymantell, line 274: I think this logic still doesn't sufficiently match what's in GetPaginated ? So the count will not match the number of results.

andymantell (changes requested):

Services/CO.CDP.OrganisationInformation.Persistence/DatabaseOrganisationRepository.cs
dbgoaco: True, the count could be different. I've updated the code to return the count as part of GetPaginated().

andymantell (approved):

dharmverma (approved):

andymantell:
{2A09C6B4-F941-4C24-8D70-730F8D41FA90}

PR #1545, opened 2025-04-30
DP-1394 Fix Recreate Onelogin while active session error fix

dharmverma:
DP-1394 Fix Recreate Onelogin while active session error fix

### DIFF
--- a/Frontend/CO.CDP.OrganisationApp/Pages/OneLogin.cshtml.cs
+++ b/Frontend/CO.CDP.OrganisationApp/Pages/OneLogin.cshtml.cs
@@ -94,6 +94,7 @@ public class OneLoginModel(
             uri += $"?redirectUri={WebUtility.UrlEncode(uriBuilder.Uri.PathAndQuery)}";
         }
 
+        session.Remove(Session.UserAuthTokens);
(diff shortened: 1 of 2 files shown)


Frontend/CO.CDP.OrganisationApp/Pages/OneLogin.cshtml.cs
```diff
@@ -94,6 +94,7 @@ private async Task<ChallengeResult> SignIn(string? redirectUri = null, string? o
             uri += $"?redirectUri={WebUtility.UrlEncode(uriBuilder.Uri.PathAndQuery)}";
         }
 
+        session.Remove(Session.UserAuthTokens);
```
dbgoaco, line 97: Why remove Session.UserAuthTokens?

dharmverma: If user is re-authenticated again, we don't want to have old auth token hanging around, they have old urn value and fails to get info from our apis in case of user has recreated the account

dbgoaco (approved):

dpatel017 (approved):

andymantell (approved):

PR #1630, opened 2025-06-16
CDPDEL-460: Forms Engine Navigation Incorrectly Redirects on Branched Path

EdwardDennis:
### Description
This pull request resolves an issue in the forms engine navigation where users could be incorrectly redirected when answering branching questions. The bug caused the system to sometimes navigate to a question on an alternative path if that question was unanswered, even if it wasn't part of the user's actual navigation route.

### Summary of Changes:
The primary issue stemmed from the GetPreviousUnansweredQuestionId logic in FormsEngine.cs not always adhering strictly to the user's actual navigation path, especially when alternative branches existed.

The fix involves the following:
**Enhanced Path Determination in FormsEngine.cs:**

- The BuildPathToQuestion method, which is utilised by GetPreviousUnansweredQuestionId, now more accurately constructs the sequence of questions the user has navigated. It correctly uses the DetermineNextQuestionIdOnPath method, which considers the user's answers to branching questions (e.g., Yes/No, FileUpload) to decide the next question on their specific path.
- This ensures that GetPreviousUnansweredQuestionId only searches for unanswered questions along the actual path taken by the user, preventing it from incorrectly identifying unanswered questions on divergent branches as blocking.

**Refined Redirect Validation in FormsQuestionPage.cshtml.cs**
- The ValidateIfNeedRedirect method in FormsQuestionPage.cshtml.cs includes a condition: if the currentQuestion is already on a path marked with FormQuestionBranchType.Alternative, the check for a previous unanswered question (and subsequent redirect) is bypassed. This complements the FormsEngine changes by ensuring that once a user is on a designated alternative path, the system respects that navigation choice.


EdwardDennis:
@andymantell @dharmverma This shouldn't affect Goaco forms but highlighting this for your awareness

PR #1634, opened 2025-06-17
CDPDEL-516: Carbon Net Zero Removal

EdwardDennis:
The Carbon Net Zero module, which was an additional supplier information module, will no longer be delivered using the forms engine. Instead, it will be delivered via an alternative method at the beginning of Q2.

This PR covers the work required to remove all existing traces of the Carbon Net Zero module from the frontend application and the database to prevent confusion and ensure a clean state for the new implementation.


andymantell (changes requested):
Is this definitely happening? It will mean schema changes for the data sharing api, which is something we try to resist if at all possible.

andymantell:
No problem @EdwardDennis - removing these entries doesn't constitute a schema change, that in itself is fine. My concern was the idea of building the carbon net zero module outside of forms - because the way you expose data in the data sharing api would then incur a schema change. If you build it inside forms - the data sharing api integration, snapshotting, etc etc all comes for "free" and with no schema change for eSenders.

PR #1680, opened 2025-06-30
DP-1574 Supplier to buyer devolved regulation content

dharmverma:
DP-1574 Supplier to buyer devolved regulation content change


MayuriGoaco (approved):
This is tested locally.Ready to merge in main

PR #1719, opened 2025-07-12
CDPDEL-601: UI Development with Mock Data

EdwardDennis:


.gitignore
```diff
@@ -47,3 +47,4 @@ msbuild.wrn
 OpenAPI
 TestResults/
 /E2ETests/appsettings.Development.json
+/.github/
```
andymantell, line 50: Why is .github ignored?

EdwardDennis: Updated thanks 👍

PR #1911, opened 2025-10-01
Fix for Welsh Modules to be optional.

tony-739573:


Frontend/CO.CDP.OrganisationApp/ShareCodeMandoryInformationService.cs
```diff
@@ -33,7 +33,8 @@ public async Task<bool> MandatorySectionsCompleted(Guid organisationId)
                 && SupplierInformationStatus.GetConnectedPersonStepStatus(supplierInfo, connectedEntities.Count) == SupplierInformationStatus.StepStatus.Completed
                 && formSections
                     .Where(s => s.Type != FormSectionType.Declaration
-                                && s.Type != FormSectionType.AdditionalSection)
+                                && s.Type != FormSectionType.AdditionalSection
```
EdwardDennis, line 36: Enhancement here could be to do the inverse check, as we're not expecting to add any more mandatory modules, but potential for new optional ones

PR #1945, opened 2025-10-17
CDPDEL-1235: Re-enable RoCT Pagination

markgorman98:


Services/CO.CDP.RegisterOfCommercialTools.WebApi/Services/CommercialToolsService.cs
```diff
@@ -48,6 +48,27 @@ public class CommercialToolsService(
...
+                totalCount = results.Count;
+            }
+        }
+        else
+        {
+            totalCount = results.Count;
+            logger.LogDebug("x-total-count header not found, using result count: {Count}", totalCount);
+        }
+
+        return (results, totalCount);
```
EdwardDennis: Here I would suggest getting the header value and then falling back to 0, using a ternary maybe. We don't really need to log this as we'd quickly know if it's broken anyway

EdwardDennis: and we could avoid mutability by handling the value directly in the totalCount assignment

EdwardDennis (approved):

PR #2024, opened 2025-11-21
feature/CDPDEL-1446 : Fix back links for buyer view

tony-739573:


Frontend/CO.CDP.OrganisationApp/Pages/Organisation/OrganisationPponSearch.cshtml.cs
```diff
@@ -259,14 +259,21 @@ public string FormatAddresses(IEnumerable<OrganisationAddress> addresses)
     };
 
     public static Shared.PaginationPartialModel CreatePaginationModel(int currentPage, int totalItems, int pageSize,
-        Guid id, string? searchText, string sortOrder)
+        Guid id, string? searchText, string sortOrder, string? origin = null)
     {
+        var url = $"/organisation/{id}/buyer/search?SearchText={Uri.EscapeDataString(searchText ?? string.Empty)}&sortOrder={sortOrder}&pageSize={pageSize}";
```
EdwardDennis: Might be worth using single string concatenation here as this code runs on every page load, search, and navigation

PR #2124, opened 2026-01-30
CDPDEL-1876 : Updated Data Protection question description text

tony-739573:


EdwardDennis (approved):
Approved - might be worth removing the corresponding Welsh until we have the new translation. It will make it easier to track and ensure the current outdated translation isn't replicated anywhere else

PR #2257, opened 2026-03-17
CDPDEL-1822: Add health check endpoint with detailed logging and application stopping notification

EdwardDennis:


Services/CO.CDP.UserManagement.Api/Program.cs
```diff
@@ -136,11 +138,48 @@
...
+
 app.UseAuthentication();
 app.UseAuthorization();
 
 app.MapControllers();
-app.MapHealthChecks("/health").AllowAnonymous();
+
+try
+{
+    using var conn = new NpgsqlConnection(connectionString);
```
jakzal, line 169: @EdwardDennis this prevents building the project without the database present.

EdwardDennis: That's the point. This was added for debugging purposes. Is this affecting anything now?

jakzal: Several people in our team complained docker builds started failing for them locally. Why is the database required at build time?

If it's only needed at runtime, the check can be wrapped with this:

```csharp
if (Assembly.GetEntryAssembly().IsRunAs("CO.CDP.UserManagement.Api"))
{
}
```

PR #2382, opened 2026-04-30
Add outbox processor for user management and update SQS queue configurations

EdwardDennis:


terragrunt/modules/ecs/templates/task-definitions/entity-verification.json.tftpl
```diff
@@ -23,12 +23,13 @@
         { "name" : "AWS_REGION", "value" : "eu-west-2"},
         { "name" : "Aws__CloudWatch__LogGroup", "value" : "${lg_name}"},
         { "name" : "Aws__CloudWatch__LogStream", "value" : "${lg_prefix}-serilog"},
-        { "name" : "Aws__SqsDispatcher__QueueUrl", "value" : "${queue_organisation_url}"},
```
webit4me, line 26: Hi @EdwardDennis 
Organisation service is struggling  to spin-up new tasks and while investigating I noticed this switch. Just checking if:
a. It is intentional 
b. SIRSI's team are aware of it

PR #2438, opened 2026-05-14
FC-955 Add fts-user-api's logs dashboard

webit4me:


webit4me:
No more backing dashboards in image

PR #2550, opened 2026-06-22
FC-1450 Enable Payments Tile in Organisation View

pablinos:


webit4me (commented):
Might be best if for the net apps we use their appsettings for non-sensitive overwrites like this

webit4me (changes requested):
I can see the prods value is already written in app settings [here](https://github.com/cabinetoffice/GCGS-Central-Digital-Platform/blob/eadd47c06c7bdd0e2316ec7c4cd2c2f9ae8005d2/Frontend/CO.CDP.OrganisationApp/appsettings.AwsDevelopment.json#L26) , it seems this value needs to go to non-prods setting and replaced by the new vlue

webit4me (approved):

PR #2602, opened 2026-07-18
Improve the docker build

pablinos:
The docker build has been taking a long time, especially when creating releases through the github workflows. On some local machines it can end up hanging, constantly restoring nuget packages.

I thought I'd have a go at fixing it. The main difference is that this builds the services in one go rather than separate images, and also removes the --no-cache from the make command. What was taking about half an hour on my machine, is now about 3-4 minutes, and it's using half the space in buildkit cache.

~~I'll keep this as a draft for now as it's the sort of change that could break things in weird and wonderful ways, but~~ it would be good to test it and see if we can get our build times under control.

This seems to be working quite well in my tests and the bottleneck is now the E2E tests.


Dockerfile
```diff
@@ -271,7 +143,7 @@ WORKDIR /app
 COPY --from=build-migrations-organisation-information /src/Services/CO.CDP.OrganisationInformation.Persistence/OrganisationInformationDatabaseMigrationConfig /app/OrganisationInformationDatabaseMigrationConfig
 COPY --from=build-migrations-organisation-information /src/Services/CO.CDP.OrganisationInformation.Persistence/StoredProcedures /app/StoredProcedures
 COPY --from=build-migrations-organisation-information /app/migrations/efbundle .
-ENTRYPOINT ["/bin/busybox", "sh", "-c", "/app/efbundle", "--connection", "Host=$OrganisationInformationDatabase__Host;Database=$OrganisationInformationDatabase__Database;Username=$OrganisationInformationDatabase__Username;Password=$OrganisationInformationDatabase__Password;"]
+ENTRYPOINT ["/bin/busybox", "sh", "-c", "/app/efbundle --connection \"Host=$OrganisationInformationDatabase__Host;Database=$OrganisationInformationDatabase__Database;Username=$OrganisationInformationDatabase__Username;Password=$OrganisationInformationDatabase__Password;\""]
```
jakzal, line 146: I'm not sure if this is a good change. `/app/efbundle --connection` might be interpreted as a command rather than a command with an option.

jakzal: Nevermind, it's passed to sh!

webit4me (approved):
love the new namespaces

PR #2644, opened 2026-08-05
FC-1917: add PDF render service secret locations

mareg:
Binds two keys in the existing cdp-<env>-fts/secrets blob: the shared API key the render worker authenticates with, and the signing certificate and key as one PEM, which is the shape the material is already held in.

The values themselves are added to the secret outside Terraform and must be in place before the service is deployed, because a task referencing a missing key fails to start. No IAM change is needed, as the execution role already reads this secret.


webit4me:
This one we can merge once those secrets are set

PR #2645, opened 2026-08-05
FC-1918: add the FTS PDF render service ECS service

mareg:
Runs the chromium PDF render service on the php cluster, which puts it on port 8070, and exposes it on the internal load balancer only: alb_enabled is false so no public listener rule is created, and internal_alb_enabled is set directly rather than through use_internal_service_urls, which is hardcoded false.

The task definition is deliberately minimal rather than a copy of the render worker's, which carries around forty variables and thirty-four secrets this service has no use for. It takes the API key and the signing PEM from the FTS secrets blob, and sets PDF_SIGNING_ALLOW_EXPIRED because the certificate in use expired in 2022 and the service refuses to start on an expired certificate otherwise. See FC-1896.

The target group health check uses the module default of /health, which the service exposes and exempts from the API key check.

Nothing points at the service yet: the render worker only calls it once PDF_RENDER_SERVICE_URL is set, which is FC-1919.

### DIFF
--- a/terragrunt/modules/ecs/locals-fts.tf
+++ b/terragrunt/modules/ecs/locals-fts.tf
@@ -321,6 +321,14 @@ locals {
     }
   )
 
+  fts_pdf_render_service_container_parameters = {
+    lg_prefix                = "app"
+    lg_region                = data.aws_region.current.region
+    pdf_render_api_key       = local.fts_secrets.pdf_render_api_key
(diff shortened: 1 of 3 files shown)


webit4me (commented):
PR is fine, but can't be merged before https://github.com/cabinetoffice/GCGS-Central-Digital-Platform/pull/2644 , which itself can't merge before setting its secrets

PR #2663, opened 2026-08-13
FC-1840  & FC-1981 add the FTS notice publish dotnet queue and worker service

vish-goaco:
Adds the fts-notice-publish-worker-dotnet ECS service.
New fts-notice-publish-dotnet FIFO queue.
Wires the dotnet queue URL and routing flag into fts web and scheduler.


terragrunt/components/root.hcl
```diff
@@ -296,6 +296,7 @@ locals {
     fts_search_indexer                   = { desired_count = 1, cpu = 4096, memory = 8192 }
     fts_user_api                         = { desired_count = 1 }
     fts_notice_publish_worker            = { desired_count = local.environment == "production" ? 5 : 3, cpu = 4096, memory = local.environment == "production" ? 16384 : 8192 }
+    fts_notice_publish_worker_dotnet     = { desired_count = 1, cpu = 4096, memory = 8192 }
```
mareg: Can we default it to `0` and only set it to 1 in `dev`?
```
desired_count = local.environment == "development" ? 1 : 0,
```

mareg (changes requested):

terragrunt/modules/ecs/templates/task-definitions/fts.json.tftpl
```diff
@@ -85,6 +86,7 @@
       { "name" : "HTTP_BASIC_AUTH_USER", "valueFrom" : "${http_basic_auth_user}"},
       { "name" : "LETSENCRYPT_KEY_AUTHORIZATION", "valueFrom" : "${letsencrypt_key_authorization}"},
       { "name" : "LETSENCRYPT_TOKEN", "valueFrom" : "${letsencrypt_token}"},
+      { "name" : "NOTICE_PUBLISH_DOTNET_ENABLED", "valueFrom" : "${notice_publish_dotnet_enabled}"},
```
mareg: **Either** confirm `NOTICE_PUBLISH_DOTNET_ENABLED` exists in the FTS secret in every environment before this applies, **or** move the line — and its definition at `locals-fts.tf:41` — to the branch that consumes it.

This is on the main FTS web task, not the new worker. ECS refuses to start a container when `valueFrom` points at a missing secret key, so getting this wrong takes down FTS itself. Nothing in FTS `main` reads the variable yet, so there's no benefit to landing it early.

terragrunt/modules/queue/main.tf
```diff
@@ -31,6 +31,18 @@ module "fts_notice_publish_queue" {
   visibility_timeout_seconds = 5 * 60
 }
 
+module "fts_notice_publish_dotnet_queue" {
+  source = "../sqs"
+
+  max_receive_count          = 3
```
mareg: Every other queue takes the module default of 10. Why is this one set to 3?

terragrunt/modules/queue/main.tf
```diff
@@ -31,6 +31,18 @@ module "fts_notice_publish_queue" {
...
+  message_retention_seconds  = 1209600 # 14 days
+  name                       = local.name_fts_notice_publish_dotnet
+  role_consumer_arn          = [var.role_ecs_task_arn]
+  role_publisher_arn         = [var.role_ecs_task_arn]
+  tags                       = var.tags
+  visibility_timeout_seconds = 5 * 60
```
mareg, line 42: Complex notices take around ten minutes in production. With a five-minute visibility timeout and no visibility extension in the .NET consumer, a slow notice is redelivered while the first attempt is still converting, up to three times. Either the consumer extends visibility while it works, or the timeout needs to clear worst-case conversion.

terragrunt/components/service/ecs/terragrunt.hcl
mareg, line 1: Run the HCL formatter and put `_url` after `queue_fts_notice_publish_url` — the block is otherwise alphabetical and column-aligned.

terragrunt/modules/ecs/templates/task-definitions/fts.json.tftpl
webit4me: It's also best if we don't define this as a secret. appsettings config in the service an env_var flag here (hopefully to be removed once out of the development phase )

terragrunt/components/root.hcl
vish-goaco: Made this change

terragrunt/modules/ecs/templates/task-definitions/fts.json.tftpl
vish-goaco: Made the change that Ali suggested - Flag becomes a dev-only env var instead of a secret.

terragrunt/modules/queue/main.tf
vish-goaco: Removed this, the dotnet queue takes the default maxReceiveCount ( 10)

terragrunt/components/service/ecs/terragrunt.hcl
vish-goaco: Made the changes - The ecs queue inputs are reordered and hclfmt-aligned.

terragrunt/modules/queue/main.tf
vish-goaco: I checked both .NET and PHP
The .NET consumer does extend visibility while it works. 
Every message is processed under a heartbeat (NoticePublishConsumer.HandleWithHeartbeatAsync / KeepMessageHiddenAsync)
It calls ChangeMessageVisibility every 30 seconds, re-hiding the message for 60 seconds each time, until the notice is processed.

This is the same mechanism the PHP worker uses today, on a queue with the same 5-minute timeout

webit4me (approved):

mareg (approved):

